PR #3, opened 2021-01-20
Dev

dhivahari-vivek-cdot:


payneBrandon:
Quality gate and code coverage badges appear to not be functional: 
image

README.md
```diff
@@ -37,6 +37,15 @@ To run this code:
...
 ```
 
+## Testing
+
+The test for the main.py script is the test_main.py script, which can be found in the /tests directory. 
+
+To run the test script:
+
+```
+python -m pytest test_main.py
```
payneBrandon, line 47: Running this errs out with the following:

image

payneBrandon:
Can we update the .gitignore file to include the standard Python .gitignore files? For instance when I try to run the test, this adds a new folder `__pycache__` and wants to add this to source control. The .gitignore should be set up to handle this case for us.

README.md
```diff
@@ -1,5 +1,52 @@
...
+## Prerequisites and Set-Up
+
+This project supports Python >= 3.5. Refer to the requirements.txt document to [pip](https://pip.pypa.io/en/stable/) install the necessary packages. The Google Cloud Storage Python packages, for example, would be installed like so:
+
+```bash
+pip install google-cloud-bigquery
+pip install google-cloud-storage
+pip install google-cloud-pubsub
+```
+
```
payneBrandon, line 29: Can we add something along the lines of "Alternatively, install all necessary tools using `pip install -r .\requirements.txt`"?

README.md
```diff
@@ -1,5 +1,52 @@
...
+## Testing
+
+The test for the main.py script is the test_main.py script, which can be found in the /tests directory. 
+
+To run the test script:
+
+```
+python -m pytest test_main.py
+```
+
```
payneBrandon, line 49: Can we add another area in here to describe what the `sample_files` directory is used for? I see the `test_main.py` calls out the the RSU-ND.json file, but it's not clear if its the one from `sample_files` directory or the one right next to the Python file. If we don't need these sample files, can we remove them?

data_manager/sample_files/gcp_test-ND.json
```diff
@@ -0,0 +1,3 @@
+{"column":1, "status":"yes"}
```
payneBrandon, line 1: This isn't a valid JSON file, are we using it anywhere? Should these objects be in a list?

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+import config
+
+def is_json_clean(rsu_data):
+    """
+    -----------------------------------------------------------------------
+    Returns TRUE if json_file is clean. FALSE otherwise.
+
+    Param: rsu_data --> RSU data as a list of dicts from raw ingest to be checked
+    -----------------------------------------------------------------------
+    """
```
payneBrandon, line 40: I like the descriptions like this where it shows up on mouse hover for the definition. Nice work!

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+    
+    #check 1: duplicate records
+    check1 = True
+    unique = []
+    for d in rsu_data:
+        if d not in unique:
+            unique.append(d)
+    if len(unique) != len(rsu_data):
+        check1 = False
+    print("\nCheck 1 is: ", check1)
```
payneBrandon, line 51: Instead of printing this as "check 1 is: false", let's get a bit more granular in our print out. We know this is a duplicate record check, so we can include that. Maybe something like "Duplicate records check failed"

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+    print(len(unique))
+    print(len(rsu_data))
+    
+    #check 2: empty records based on timeReceived key
+    check2 = True
+    for d in rsu_data:
+        if len(d["timeReceived"]) == 0:
+            check2 = False
+            break
+    print("\nCheck 2 is : ", check2)
```
payneBrandon, line 61: Same as above, let's print out some details about this "Check 2" and what it was that we were checking for.

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+
+    return check
+
+def rsu_raw_bucket(client, filename, filepath, bucket_name):
+    """
+    -----------------------------------------------------------------------
+    Executes the transfer of raw data from the roadside unit to the 
+    rsu_raw-ingest bucket in the GCP which stores raw, unclean data.
+
+    Param: client --> object referencing GCP Storage/Bucket Client
```
payneBrandon, line 76: Can we add descriptions for each of the params?

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+    current_time = datetime.datetime.utcnow()                   
+    log_message = Template('Raw ingest updated with new file at $time')
+    logging.info(log_message.safe_substitute(time=current_time))
+    
+    print("End of bucket #1.")
+
+def help_data_lake(list_blobs, raw_bucket, lake_bucket):
+    """
+    -----------------------------------------------------------------------
+    Helper function for the rsu_data_lake_bucket() function which
```
payneBrandon, line 94: Can we clean up this description a bit and include that this function takes in a list of blobs, checks if they are "json clean", and then copies them to the provided lake_bucket?

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+    to the data lake bucket in the GCP to be cleaned/filtered/aggregated/etc.
+
+    Param: client --> object referencing GCP Storage/Bucket Client
+    -----------------------------------------------------------------------
+    """
+    print("Beginning of bucket #2.")
+    
+    raw_bucket = client.get_bucket(r_bucket)            # source bucket
+    lake_bucket = client.get_bucket(l_bucket)           # destination bucket
+    raw_blobs = client.list_blobs(raw_bucket)
```
payneBrandon, line 121: Are we maintaining a list of completed blobs somewhere? This would appear to list all blobs in the raw_bucket, then copy them to the data lake. If the function were to run multiple times would we be duplicating data in the data lake?

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+        future = client.publish(topic,data_string)       # when message is published, the client returns a "future"
+        print(future.result())       
+
+def rsu_data_warehouse_bucket(pub_client, storage_client, topic, bucket):
+    """
+    -----------------------------------------------------------------------
+    Will push data from the data lake to the designated Pub/Sub topic which
+    will serve as a form of short-term data warehouse.
+
+    Param: b_client --> object referencing GCP Storage/Bucket Client
```
payneBrandon, line 156: Can we update these to reflect the params passed in?

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+    Main function.
+    Triggered from a message on a Cloud Pub/Sub topic.
+
+    Param - data: (dict) Event payload.
+    Param - context: (google.cloud.functions.Context) Event metadata.
+    -----------------------------------------------------------------------
+    """
+
+    # setting the Google Application Credentials to JSON with service account key
+    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = r"CDOT CV ODE Dev-4d9416c81201.json"
```
payneBrandon, line 184: Let's add some text to the readme that will talk to this json file. What it does, what format to use for it, and that it's been excluded in the .gitignore due to holding sensitive connection information.

data_manager/source_code/main.py
```diff
@@ -0,0 +1,204 @@
...
+    print("\nLoaded Google App credentials.")
+
+    json_file = 'RSU-ND-clean.json'
+
+    raw_ingest_bucket = 'rsu_raw-ingest'
+    data_lake_bucket = 'rsu_data-lake'
```
payneBrandon, line 191: Are some of these values ones we could pass into the `main` function? This would enable us to run against various environments easier. For instance these buckets may be named the same across environments, but the topic `cdot-cv-ode-dev` below will likely be renamed based on environment.

sonar-project.properties
```diff
@@ -10,7 +10,10 @@ sonar.organization=cdot-cv
 sonar.links.homepage=https://github.com/CDOT-CV/RSU_Management
 sonar.links.scm=https://github.com/CDOT-CV/RSU_Management
 sonar.links.issue=https://github.com/CDOT-CV/RSU_Management/issues
+sonar.python.coverage.reportPaths=*coverage*.xml
 # Exclusions
+# sonar.exclusions=tests/*
```
payneBrandon, line 15: Can we delete this line?

tests/test_main.py
```diff
@@ -0,0 +1,101 @@
...
+import pytest
+from google.cloud.storage import Client
+from google.cloud.pubsub_v1 import PublisherClient
+
+
+"""
+This script handles the unit testing for main.py.
+"""
+
+def test_is_json_clean():
```
payneBrandon, line 18: Each unit test should try to `assert` a single case, testing a single "unit". For this test, let's break it into three tests - each named for what they are testing. 

IE for the first one we can test `def test_is_json_clean_true` and then test only the succeeding json. Then the second test will be called `def test_is_json_clean_false_duplicate_records` or similar to describe what is being tested and expected.

sonar-project.properties
dhivahari-vivek-cdot: I thought we put this line in here so that the lines in test_main.py weren't included in the code coverage? Before I put this exclusion in, test_main.py was included in the code coverage, and I saw that the Work Zone repo had the same exclusion line.

data_manager/source_code/main.py
dhivahari-vivek-cdot: So this (along with most of your suggestions) came because I refactored a lot of the code to make it easier to test and forgot to update accordingly. I'm sure I can pass the project name cdot-cv-ode-dev into main, though. Or maybe in the next iteration, have the user enter in the project name and JSON credential path?

sonar-project.properties
payneBrandon: Isn't this line commented out? The line directly below it tells sonar to exclude coverage.xml files and everything under `tests` directory. If its a commented out line we should remove it.

data_manager/source_code/main.py
dhivahari-vivek-cdot: This was one of the concerns I had as well. Basically, I didn't know if you wanted to delete the blobs from the raw ingest once it was verified for cleanliness and passed into the data lake or keep them around for some reason. If so, deleting the blob from the raw ingest I feel would be a good way to ensure we don't duplicate.

sonar-project.properties
dhivahari-vivek-cdot: Whoopsie. You're right. I'll remove that now.

data_manager/source_code/main.py
payneBrandon: I'm fine with leaving it hard-coded for this PR as long as we address it for the next one. Going into the GCP we'll want the flexibility of changing those variables without making code changes.

sonar-project.properties
dhivahari-vivek-cdot: Read the wrong line.

data_manager/source_code/main.py
dhivahari-vivek-cdot: Sounds good. I'll be sure to add it to my user story for next sprint. This type of refactoring will likely come up a lot once I start moving these scripts to the GCP, so I think that's a good idea.

data_manager/source_code/main.py
dhivahari-vivek-cdot: Is that okay with you? Or did you have something else in mind for raw ingest blobs that have already been looked at?

payneBrandon: @TonyEnglish do you think we need to keep the raw blobs around for historical purposes or can we remove? Alternatively we may be able to refactor the code to watch these buckets and trigger the various functions each time a new blob is added to the watched bucket. Thoughts?

tests/test_main.py
dhivahari-vivek-cdot: Updated test_main to reflect these changes.

data_manager/source_code/main.py
```diff
@@ -0,0 +1,209 @@
...
+    Param: list_blobs --> the data to be stored in Pub/Sub
+    Param: topic --> the Pub/Sub topic designated as the data warehouse
+    -----------------------------------------------------------------------
+    """
+    for blob in list_blobs:
+        data_string = blob.download_as_string()                 # data MUST be a byte string
+        future = client.publish(topic,data_string)       # when message is published, the client returns a "future"
+        print(future.result())       
+
+def rsu_data_warehouse_bucket(pub_client, storage_client, topic, bucket):
```
payneBrandon, line 150: Missing param description for `bucket` here

README.md
payneBrandon: I'm still getting an error when trying to run the test.

dhivahari-vivek-cdot: Oh. So, if you're running the test on your local computer, you may have to rearrange a few things. I had to rewrite some of the code so that it would pass the sonarcloud tests. I'm guessing I should probably write about this in the README? Basically, if running on local computer:

1. make sure main.py and test_main.py are in the same folder.
2. comment out line 3 (from data_manager.source_code import main)
3. add the following to the list of imports: 
from main import rsu_raw_bucket
from main import rsu_data_warehouse_bucket
from main import rsu_data_lake_bucket
from main import help_warehouse
from main import help_data_lake
from main import is_json_clean
from main import main
4. anytime a function from main.py is called throughout the test script, remove the " main. " in front of it. So, instead of main.is_json_clean(mock_json), just write is_json_clean(mock_json).
5. at the end, call each test function without parameters, like so:
test_is_json_clean()
test_rsu_raw_bucket()
test_rsu_data_lake_bucket()
test_rsu_data_warehouse_bucket()
test_help_warehouse()
test_help_data_lake()
test_main()

Jacob and I went through this part together, and we had to rearrange this script depending on if we ran it locally vs. if we wanted it to pass sonarcloud.

data_manager/source_code/main.py
dhivahari-vivek-cdot: Updated this description.

README.md
payneBrandon: That's a bit excessive to get unit tests to run. It looks like `mock` is a module you can install with pip. If you add `mock` to the requirements.txt it will install it when running the initial `pip install -r requirements.txt` and then the unit tests succeed.

dhivahari-vivek-cdot: ...Oh. That does sound much better. Added mock to the requirements.txt. Thank you.

dhivahari-vivek-cdot:
> Can we update the .gitignore file to include the standard Python .gitignore files? For instance when I try to run the test, this adds a new folder `__pycache__` and wants to add this to source control. The .gitignore should be set up to handle this case for us.

data_manager/source_code/main.py
TonyEnglish: I think the raw_bucket would keep these data sets, the lake_bucket would retain the clean files, then the data would move into pub sub for initial reporting (raw), then to standard pub sub for a transformed data set (matching across RSU vendors), then possible one more for aggregate data (like daily counts). I would expect the data warehouse would collect standard and aggregate data  only.

Each data set would have a life time for usefulness and purged after that date, Huy and Ashley would need to weigh in on this, but i would propose around a month for raw, 2 months for data lake, a month for each of the pub sub topics, and the data warehouse would be for a couple years

payneBrandon: Resolving for now, but @dhivahari-vivek-cdot let's circle back on this in our sprint kickoff tomorrow.

payneBrandon (approved):
Looks good!

payneBrandon (approved):

PR #19, opened 2021-02-13
Dev to main (v2)

dhivahari-vivek-cdot:
Closed my other PR and submitting this one instead, since the quality gate passed and I have a ~90.1% coverage now. Submitted unit tests for all three GCP functions - located in the tests/ dir. Monday, though, I want to take another stab at the code lines I couldn't cover.


GCP_cloud_functions/lake_to_data_warehouse/lake_to_data_warehouse.py
```diff
@@ -1,41 +1,43 @@
...
-        log_message = Template('Published message to the warehouse pub/sub topic at $time')
-        logging.info(log_message.safe_substitute(time=current_time))
-
-    except Exception as error:
-        log_message = Template('Failed to perform actions with data warehouse Pub/Sub topic due to $message')
-        logging.error(log_message.safe_substitute(message=error))
-
-    
+import sys
+sys.path.append("..")
```
payneBrandon, line 2: What are we gaining by importing `sys` and appending to the path?

GCP_cloud_functions/raw_to_data_lake/raw_to_data_lake.py
```diff
@@ -1,71 +1,73 @@
...
-            current_time = datetime.datetime.now()
-            log_message = Template('Data lake updated at $time')
-            logging.info(log_message.safe_substitute(time=current_time))
-
-    except Exception as error:
-        log_message = Template('Failed to perform operations on raw and/or data lake storage buckets due to $message')
-        logging.error(log_message.safe_substitute(message=error))
-  
-
+import sys
```
payneBrandon, line 1: Same as before, are we gaining anything by importing sys and appending to the path? I don't see that we're looking at any path here.

GCP_cloud_functions/rsu_to_raw_ingest/rsu_to_raw_ingest.py
```diff
@@ -1,39 +1,41 @@
...
-        log_message = Template('Raw ingest uploaded with new data at $time')
-        logging.info(log_message.safe_substitute(time=current_time))
-
-    except Exception as error:
-        log_message = Template('Failed to perform operations on raw ingest storage bucket due to $message.')
-        logging.error(log_message.safe_substitute(message=error))
-            
-
+import sys
+sys.path.append("..")
```
payneBrandon, line 2: Are we using this to help with the `upload_from_filename` below?

README.md
```diff
@@ -56,6 +56,22 @@ To run this code on a local machine:
 python3 main.py
 ```
 
+## Google Cloud Storage (GCS): Cloud Function Set-Up
+
+The modularized code in main.py can be easily refactored into smaller modules, or Cloud Functions, in the GCP. CDOT's implementation divided main.py into three Cloud Functions, and the GCP_cloud_functions directory contains the files necessary for each GCP Cloud Function set-up. For instance, the GCP_cloud_functions/rsu-to-raw-ingest folder contains every file needed to set up and deploy the rsu-to-raw-ingest Cloud Function (which will pull new data from the RSU and send it to the data bucket containing the raw ingest). Additionally, the config.py file contains the storage/container identifiers used in each Cloud Function, and must be included in each individual Cloud Function deployment.
```
payneBrandon, line 61: We'll want to use environment variables rather than importing this `config.py` file. We'll also need to document in here what variables are expected to be provided (ie we need data lake, raw injest, etc. or else the function will bomb)

README.md
```diff
@@ -56,6 +56,22 @@ To run this code on a local machine:
...
+
+CDOT's Cloud Function set-up refactors main.py into three Cloud Functions: 
+
+- **rsu-to-raw-ingest** function: retrieves the raw data ingest from the RSU(s) and sends it to the designated data bucket in the GCS which stores the raw ingest. This function is triggered by a Pub/Sub topic receiving timely messages from the Cloud Scheduler. For instance, the Cloud Scheduler may publish a message to this Pub/Sub topic every five minutes, triggering the Cloud Function to pull from the RSU and send to the designated data bucket every five minutes.
+- **raw-to-data-lake** function: retrives new uploads from the data ingest bucket and "checks" its cleanliness before sending approved, "clean" data to the designated "data lake" storage bucket in the GCS. This function is triggered by any new data upload to the data ingest storage bucket.
```
payneBrandon, line 66: Let's add a bit here to what we are considering "cleanliness". Talk to what the function is actually doing when it's checking, and what happens if that check fails

tests/test_rsu_to_raw_ingest.py
```diff
@@ -0,0 +1,48 @@
+import sys
+sys.path.append("..")
+from GCP_cloud_functions.rsu_to_raw_ingest import rsu_to_raw_ingest
+import mock
+import base64
+import datetime
+from unittest.mock import patch, mock_open
```
payneBrandon, line 7: Lines 7 - 10 don't seem to be used. We're adding the client as a mock, but it's coming through as an anonymous type. We can either specify a type for that parameter, or remove the import.

GCP_cloud_functions/lake_to_data_warehouse/lake_to_data_warehouse.py
dhivahari-vivek-cdot: I believe this was so that it could access the correct file in the github repo. I used this during the last test and I believe both Abinash and I were advised to do this in our repos.

dhivahari-vivek-cdot: But I can go ahead and remove it to see if it still goes to the right dir.

GCP_cloud_functions/raw_to_data_lake/raw_to_data_lake.py
dhivahari-vivek-cdot: See above comment!

GCP_cloud_functions/lake_to_data_warehouse/lake_to_data_warehouse.py
dhivahari-vivek-cdot: -sigh- You were right. Removed the sys.append!

GCP_cloud_functions/raw_to_data_lake/raw_to_data_lake.py
dhivahari-vivek-cdot: Fixed.

GCP_cloud_functions/rsu_to_raw_ingest/rsu_to_raw_ingest.py
dhivahari-vivek-cdot: Fixed.

tests/test_rsu_to_raw_ingest.py
dhivahari-vivek-cdot: Fixed for all files.

README.md
dhivahari-vivek-cdot: Fixed.

tests/test_rsu_to_raw_ingest.py
payneBrandon: Not fixed, still importing several unused values. `patch`, `mock_open`, and `TestCase` all appear to be unused imports.

GCP_cloud_functions/lake_to_data_warehouse/lake_to_data_warehouse.py
payneBrandon: This is still in the `lake_to_data_warehouse.py` file...

GCP_cloud_functions/raw_to_data_lake/raw_to_data_lake.py
payneBrandon: Still seeing it, did you remove?

GCP_cloud_functions/lake_to_data_warehouse/lake_to_data_warehouse.py
dhivahari-vivek-cdot: Whoops, forgot about the actual cloud code. Went through and removed sys.append to all the source code files as well. Should have covered it.

GCP_cloud_functions/raw_to_data_lake/raw_to_data_lake.py
dhivahari-vivek-cdot: Yup, sorry about that. Should be fixed now.

tests/test_rsu_to_raw_ingest.py
dhivahari-vivek-cdot: Sorry about that, the unit code went through a few changes before I pushed to the repo and forgot to update the packages here. Thanks for catching that, should be updated now!

dhivahari-vivek-cdot: Interesting, though, the code coverage dropped from 90.2 to 89.8 after doing this...hm.

dhivahari-vivek-cdot: Oh wait, that makes sense, since there are fewer lines of code. nvm! i'll keep going at it w/ code coverage!

tests/test_rsu_to_raw_ingest.py
```diff
@@ -0,0 +1,48 @@
...
+mock_context.timestamp = '2019-07-15T22:09:03.761Z'
+
+@mock.patch("google.cloud.storage.Client")
+def test_PubSubmitted(client):
+    
+    name = 'Testing GCP automation, 1-2-3!'
+    data = {'data': base64.b64encode(name.encode())}
+    raw_bucket = client().get_bucket
+    rsu_to_raw_ingest.rsu_to_raw_ingest(data, mock_context)
+    blob = raw_bucket().blob
```
payneBrandon, line 22: I see several places where just `assert True` is being called. We're not really asserting anything useful here, just a bare boolean value. Is there a reason to keep these assertions?

dhivahari-vivek-cdot: Removed those statements!

tests/test_rsu_to_raw_ingest.py
```diff
@@ -0,0 +1,48 @@
...
+import pytest
+
+"""
+This script handles the unit testing for the rsu-to-raw-ingest Cloud Function.
+"""
+
+mock_context = mock.Mock()
+mock_context.event_id = '617187464135194'
+mock_context.timestamp = '2019-07-15T22:09:03.761Z'
+
```
payneBrandon, line 14: We'll want to also assert that our `get_bucket` was called with the expected config value. Inject the config along with the Client above to set this up

tests/test_rsu_to_raw_ingest.py
```diff
@@ -0,0 +1,48 @@
...
+@mock.patch("google.cloud.storage.Client")
+def test_PubSubmitted(client):
+    
+    name = 'Testing GCP automation, 1-2-3!'
+    data = {'data': base64.b64encode(name.encode())}
+    raw_bucket = client().get_bucket
+    rsu_to_raw_ingest.rsu_to_raw_ingest(data, mock_context)
+    blob = raw_bucket().blob
+    blob_name = (str(datetime.datetime.now().strftime("%Y-%m-%d %H:%M")))
```
payneBrandon, line 23: This test method doesn't seem to be set up correctly. Several of the functions from above have been turned into properties and aren't working correctly.

We'll also want to raise an exception, not just test for one. Look into `side_effect` and how to use that to raise an exception when the `blob()` function is called.

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,81 @@
...
...eAlerts': '', 'description': '', 'trailers': '', 'classification': '', 'classDetails': '', 'vehicleData': '', 'weatherReport': '', 'weatherProbe': '', 'obstacleDetection': '', 'disabledVehicle': '', 'speedProfile': '', 'rtcm': '', 'message': '{"basicsafetymessage":{"coredata":{"accelset":{"lat":2001,"long":0,"vert":-127,"yaw":0},"accuracy":{"orientation":8100,"semimajor":40,"semiminor":40},"angle":127,"brakes":{"abs":{"value":2},"auxbrakes":{"value":0},"brakeboost":{"value":0},"scs":{"value":2},"traction":{"value":2},"wheelbrakes":[false,false,false,false,false]},"elev":25313,"heading":19072,"id":"j6EXKQ==","lat":397452755,"long":-1056791004,"msgcnt":91,"secmark":6500,"size":{"length":1020,"width":260},"speed":1506,"transmission":{"value":2}},"partii":[{"vehiclesafetyextensions":{"pathhistory":{"crumbdata":[{"elevationoffset":11,"latoffset":-4120,"lonoffset":-9250,"timeoffset":300},{"elevationoffset":41,"latoffset":-12864,"lonoffset":-30954,"timeoffset":980}]},"pathprediction":{"confidence":180,"radiusofcurve":-12880}}},{"partii_id":1,"specialvehicleextensions":{"vehiclealerts":{"lightsuse":{"value":1},"multi":{"value":0},"sirenuse":{"value":1},"ssprights":0}}}]},"messageid":20}'},
```
payneBrandon, line 16: Let's add in some edge case testing here as well. What  happens if we pass in an empty string? What about poorly formatted JSON?

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,81 @@
...
+    event = {
+        'bucket': 'rsu_raw-ingest',
+        'name': 'test',
+        'metageneration': 'some-metageneration',
+        'timeCreated': '0',
+        'updated': '0'
+    }
+    context = mock.MagicMock()
+    context.event_id = 'some-id'
+    context.event_type = 'gcs-event'
```
payneBrandon, line 64: This test for raising an Exception is hiding an issue in the testing code. The exception being raised is `AssertionError: expected call not found`. This is not validating that an exception is being raised in the tested code, but rather saying the `blob()` method was never called at all.

We'll want to explicitly inject our exceptions and verify they are getting raised as expected. For instance, `get_bucket` can raise "google.cloud.exceptions.NotFound" exception. We should inject his and verify it gets raised.

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,81 @@
...
+
+    raw_bucket = 'rsu_raw-ingest'
+    raw_bucketOBJ = client().get_bucket(raw_bucket)
+    raw_blob = raw_bucketOBJ.blob(event['name'])
+
+    raw_blob.upload_from_string('{"timeReceived": "2020-05-14T11:37:06Z", "year": "2020", "month": "05", "day": "14", "hour": "11", "version": "1.1.0", "type": "bsm"}')
+    lake_bucket = 'rsu_data-lake'
```
payneBrandon, line 47: We'll want another test for what happens if the `blob()` function returns None (which we'll need to add a check for in the raw_to_data_lake code)

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,81 @@
...
+        'timeCreated': '0',
+        'updated': '0'
+    }
+    context = mock.MagicMock()
+    context.event_id = 'some-id'
+    context.event_type = 'gcs-event'
+
+    raw_bucket = 'rsu_raw-ingest'
+    raw_bucketOBJ = client().get_bucket(raw_bucket)
+    raw_blob = raw_bucketOBJ.blob(event['name'])
```
payneBrandon, line 44: We haven't validated anything in this test, just made a few calls to the mocked object in this test method, then called our unit under test.

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,61 @@
...
+@mock.patch("google.cloud.pubsub_v1.PublisherClient")
+@mock.patch("google.cloud.storage.Client")
+def test_Success(client, publish_client):
+
+    event = {
+        'bucket': 'rsu_data-lake',
+        'name': 'test',
+        'metageneration': 'some-metageneration',
+        'timeCreated': '0',
+        'updated': '0'
```
payneBrandon, line 14: I'm not sure what we're doing with the lines 20-23 in here. These are just making calls against the mocked instance right? But we're not asserting anything or setting anything.

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,61 @@
...
+    }
+    context = mock.MagicMock()
+    context.event_id = 'some-id'
+    context.event_type = 'gcs-event'
+
+    lake_bucket = 'rsu_lake-bucket'
+    lake_bucketOBJ = client().get_bucket(lake_bucket)
+    lake_blob = lake_bucketOBJ.blob(event['name'])
```
payneBrandon, line 22: There's a few other things in here we should verify happened. We'll want to inject the config file then verify that we're calling out to `publisher.topicPath` with the expected parameters. We'll also want to check we're passing proper parameters to `get_bucket` and `get_blob` methods.

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,61 @@
...
+        'bucket': 'rsu_data-lake',
+        'name': 'test',
+        'metageneration': 'some-metageneration',
+        'timeCreated': '0',
+        'updated': '0'
+    }
+    context = mock.MagicMock()
+    context.event_id = 'some-id'
+    context.event_type = 'gcs-event'
+
```
payneBrandon, line 44: Again, we're not really testing our method's ability to handle the exception here, just that the method is never called.

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,61 @@
...
+    lake_bucket = 'rsu_lake-bucket'
+    lake_bucketOBJ = client().get_bucket(lake_bucket)
+    lake_blob = lake_bucketOBJ.blob(event['name'])
+    lake_blob.upload_from_string('{"timeReceived": "2020-05-14T11:37:06Z", "year": "2020", "month": "05", "day": "14", "hour": "11", "version": "1.1.0", "type": "bsm"}')
+    topic_path = publish_client().topic_path('cdot-cv-ode-dev','rsu_data_warehouse')
+
+    lake_to_data_warehouse.rsu_data_warehouse_bucket(event, context)
```
payneBrandon, line 26: We'll want to cover the case that the `get_bucket` method throws an error, as well as if `blob()` returns None

tests/test_rsu_to_raw_ingest.py
dhivahari-vivek-cdot: Fixed!

tests/test_rsu_to_raw_ingest.py
payneBrandon: I still see several `assert True` statements in the test_main.py

dhivahari-vivek-cdot: Fixed.

tests/test_rsu_to_raw_ingest.py
payneBrandon: I don't see where you're asserting this? I'd expect something like `client().get_bucket.assert_called_with(...)`

tests/test_rsu_to_raw_ingest.py
dhivahari-vivek-cdot: Note for all exception-related testing: Spoke with Drew. He doesn't seem to find anything wrong with the syntax of my unit test for this exception, but he did say that if the tests are finishing before they throw an error then that would explain why you aren’t getting an error. Even though that exception has been thrown correctly in the past (such as when I accidentally forgot to use the right bucket name). So...I am still a little stumped here, but I'll keep going. There's another patch of code in the raw-to-data-lake.py (the "if is_json_true()" block) that I'll aim to cover to go above my current 89.2% coverage. 

He also told me that if I could get my coverage above 90% without the exceptions that I could just go without full coverage on the exception.

tests/test_rsu_to_raw_ingest.py
payneBrandon: Ah, looks like it was just pushed up after this comment

dhivahari-vivek-cdot: Yup! A bit of a delay there.

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: I've got the empty string case pushed in, but I wanted to get to >90% code coverage, so I'm working on the aforementioned is_json_clean if statement code block to assert that properly and cover this entire code base.

tests/test_rsu_to_raw_ingest.py
payneBrandon: Rather than rely on the `config.py` file to be in the GCP_cloud_functions directory, where we don't have control over the variables and they may change which would invalidate this test, let's inject a config with controlled variables.

dhivahari-vivek-cdot: Currently the GCP set-up has each function packaged separately with its own config.py file --> this was how I believe the GCP documentation showed us we could include these variables. Do we want to completely change that set-up in the original code?

tests/test_main.py
```diff
@@ -94,8 +86,6 @@ def test_help_warehouse(client, publish_client):
     topic_path = publish_client().topic_path('cdot-cv-ode-dev','rsu_data_warehouse')
     main.help_warehouse([lake_blob], publish_client(),topic_path)
     publish_client().publish.assert_called_with(topic_path, client().get_bucket().blob().download_as_string())
-    
-    assert True
 
 def test_main():
```
payneBrandon, line 90: We shouldn't be just asserting True in this case. The main function has numerous actions going on that we could test are getting fired off, for instance the `rsu_raw_bucket()` was called with known values, as was the data lake and warehouse methods. We can assert they were called properly.

dhivahari-vivek-cdot: I believe I just changed this. Maybe the pushes are delayed>

tests/test_rsu_to_raw_ingest.py
payneBrandon: No, that is fine for running the functions. But for testing them, we want control over the variables. This unit test is saying "I have a file in my local directory that I'm relying on existing and never changing the values of". For a unit test, we should be controlling for that case, mocking the config values and injecting test values. Then we can test knowing our exact values should always be in there.

tests/test_main.py
payneBrandon: No, it's still there in this test_main() method: https://github.com/CDOT-CV/RSU_Management/blob/9c671d3f1364539d93ecb444056ff65752e50409/tests/test_main.py#L96

tests/test_rsu_to_raw_ingest.py
payneBrandon: If we need to roll this unit testing task over to the next sprint we can do that as well. The coverage percentage is a little misleading here. We want to cover the case that an exception is thrown so we know the code handles this.

tests/test_main.py
dhivahari-vivek-cdot: Oh, I see what you mean. One moment.

dhivahari-vivek-cdot: Sorry about that. I didn't focus on the test_main because I thought I was to focus on the unit tests for the GCP functions (since you had approved the test_main last sprint). I may have misunderstood this, would have looked at test_main otherwise!

tests/test_rsu_to_raw_ingest.py
dhivahari-vivek-cdot: Hmm. I think I may have to do that. I perhaps carelessly thought I was finished since I hit that 90.2% on Friday - and why I wasn't as worried about covering the Exceptions that early in the sprint. Could we push this into the next sprint?

payneBrandon: Go ahead and add a comment to your unit testing task saying that we did not finish in this sprint and are rolling to the next one to ensure unit tests are covering edge cases, then we can move it to the next sprint.

tests/test_main.py
payneBrandon: I don't think I dug deep enough into the unit tests last sprint to get all the coverage in them we want. In any case, if we need to roll this into the next sprint with the above unit test roll over we can. We'll just need to comment in the task about what needs done.

tests/test_rsu_to_raw_ingest.py
dhivahari-vivek-cdot: Sounds good. I also made some pushes to cover some of the missed lines but because Travis has halted, I don't think it's updated the coverage in SonarCloud (passed locally on my machine)? So, the coverage might be slightly higher than 89.8? Unsure how to proceed with this new development.

tests/test_main.py
dhivahari-vivek-cdot: Totally makes sense, and I have made that comment in my story!

tests/test_rsu_to_raw_ingest.py
payneBrandon: Are we able to mock that config object so we control the values?

dhivahari-vivek-cdot: Oh! Skip the rsu_to_raw_ingest testing for now, since that will include the Panasonic stuff and will come with our next PR. Maybe I should delete this file instead?

tests/test_rsu_to_raw_ingest.py
dhivahari-vivek-cdot: I've covered the Exception testing for the latter two Cloud Functions under Drew's guidance. For this, again, this function's updated script and tests will come next PR (from story branch CV-66 to dev).

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: I think I've updated this?

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: I believe I've done this under Drew's guidance, please take a look!

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: I believe I have done so as mentioned above!

tests/test_lake_to_data_warehouse.py
dhivahari-vivek-cdot: Should have this fixed.

tests/test_lake_to_data_warehouse.py
dhivahari-vivek-cdot: Should be fixed.

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,80 @@
...
+     
+    assert raw_to_data_lake.is_json_clean(mock_json2) is False       # missing records = CHECK SHOULD FAIL
+
+def test_is_json_empty():
```
payneBrandon, line 31: Can we rename to include the method being tested? Maybe `test_is_json_clean_false_empty` to follow your other methods naming convention?

GCP_cloud_functions/lake_to_data_warehouse/lake_to_data_warehouse.py
payneBrandon: This seems to have crept back into the `test_main.py` file.

payneBrandon:
There's several unused imports in the test files, can we clean those up?

payneBrandon:
It looks like we've removed the config usage, can we remove the import in `main.py` and remove the file? As a side note here, there seems to be other unused imports in that `main.py` file we can remove as well.

README.md
payneBrandon: Can we add a section to this readme to explain what environment variables are expected?

tests/test_main.py
```diff
@@ -45,8 +45,6 @@ def test_rsu_raw_bucket(client):
     blob.assert_called_with(file_name)
```
payneBrandon, line 45: This is a good "perfect case" test scenario. We'll want to add a test showing that we expect it to throw an exception if the bucket doesn't exist for instance.

tests/test_main.py
```diff
@@ -56,8 +54,6 @@ def test_rsu_data_lake_bucket(client):
     main.rsu_data_lake_bucket(client(), raw_bucket, data_lake_bucket)
     client().list_blobs.assert_called_with(raw_bucketOBJ)
 
-    assert True
-
 @mock.patch("google.cloud.storage.Client")
 def test_help_data_lake(client):
```
payneBrandon, line 58: We're not asserting anything in our test with this one, just making calls. We should mock the `is_json_clean` method and have a test showing our expected behavior when that returns true, and when it returns false.

payneBrandon: Also, can we name this helper function something that describes what it's doing? Maybe `verify_and_copy_blobs_to_data_lake` or something similar.

tests/test_main.py
```diff
@@ -45,8 +45,6 @@ def test_rsu_raw_bucket(client):
     blob.assert_called_with(file_name)
     blob().upload_from_filename.assert_called_with(filename=file_path)
 
-    assert True
-
 @mock.patch("google.cloud.storage.Client")
 def test_rsu_data_lake_bucket(client):
```
payneBrandon, line 49: If we mock the helper method and inject it here, we can verify it was called with our expected parameters.

tests/test_main.py
```diff
@@ -56,8 +54,6 @@ def test_rsu_data_lake_bucket(client):
     main.rsu_data_lake_bucket(client(), raw_bucket, data_lake_bucket)
```
payneBrandon, line 54: We'll want another test showing that we expect this method to throw an exception if the buckets do not exist.

tests/test_main.py
```diff
@@ -81,8 +75,6 @@ def test_rsu_data_warehouse_bucket(client, publish_client):
     main.rsu_data_warehouse_bucket(publish_client(), client(), topic_path, data_lake_bucket)
     client().list_blobs.assert_called_with(lake_bucketOBJ)
 
-    assert True
-
 @mock.patch("google.cloud.pubsub_v1.PublisherClient")
 @mock.patch("google.cloud.storage.Client")
 def test_help_warehouse(client, publish_client):
```
payneBrandon, line 80: Again if we mock the helper method (`help_warehouse`), we can verify it was called with our expected parameters.

tests/test_main.py
```diff
@@ -94,8 +86,6 @@ def test_help_warehouse(client, publish_client):
     topic_path = publish_client().topic_path('cdot-cv-ode-dev','rsu_data_warehouse')
     main.help_warehouse([lake_blob], publish_client(),topic_path)
```
payneBrandon, line 87: We'll want a test showing we expect an exception if the get_bucket method fails.

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: Fixed!

tests/test_main.py
dhivahari-vivek-cdot: Fixed!

tests/test_main.py
dhivahari-vivek-cdot: Fixed.

tests/test_main.py
dhivahari-vivek-cdot: Fixed.

tests/test_main.py
dhivahari-vivek-cdot: Fixed.

tests/test_main.py
dhivahari-vivek-cdot: Fixed and fixed.

tests/test_main.py
dhivahari-vivek-cdot: Fixed.

tests/test_main.py
payneBrandon: This helper method was added to the `test_rsu_data_lake_bucket_ExceptionRaised_BucketNotFound` and verified to not have been called which is great, but we can add it to this method and verify it WAS called as well.

tests/test_main.py
```diff
@@ -1,15 +1,11 @@
-import sys
-sys.path.append("..")
 from data_manager.source_code import main
 import mock
+import base64
```
payneBrandon, line 3: Several of these imports are not used, can we remove?

data_manager/source_code/main.py
```diff
@@ -4,80 +4,63 @@
...
-data warehouse. The raw ingest and data lake at present are located in their respective
-buckets in the GCP. The data warehouse is currently a Pub/Sub topic to which cleaned 
-data will be published in byte string form.
+data warehouse. Furthermore, this script is one meant to be run locally, and was mainly designed
+to demonstrate the data flow process between storage containers in the GCP. 
+The raw ingest and data lake at present are located in their respective buckets in the GCP. 
+The data warehouse is currently a Pub/Sub topic to which cleaned data will be published in byte string form.
 """
 
 import json
```
payneBrandon, line 13: Several of these imports are unused, can we remove?

tests/test_main.py
```diff
@@ -56,20 +66,45 @@ def test_rsu_data_lake_bucket(client):
...
+    data_lake_bucket = client().get_bucket
+    client().get_bucket.side_effect = exceptions.NotFound('testing')
+
+    with pytest.raises(exceptions.NotFound):
+        main.rsu_data_lake_bucket(client(), raw_bucket, data_lake_bucket)
+    
+    assert not mockHelpDataLake.called
 
 @mock.patch("google.cloud.storage.Client")
 def test_help_data_lake(client):
```
payneBrandon, line 83: Since the function has been renamed, can we rename the test accordingly?

tests/test_main.py
```diff
@@ -56,20 +66,45 @@ def test_rsu_data_lake_bucket(client):
...
+
+    raw_blob.upload_from_string('{"timeReceived": "2020-05-14T11:37:06Z", "year": "2020", "month": "05", "day": "14", "hour": "11", "version": "1.1.0", "type": "bsm"}')
+    raw_blob.download_as_bytes.return_value = json_bytes
 
-    assert True
+    main.verify_and_copy_blobs_to_data_lake([raw_blob], raw_bucket, lake_bucket)
+    raw_bucket.copy_blob.assert_called_with(raw_blob, lake_bucket)
+
+@mock.patch("google.cloud.storage.Client")
+def test_help_data_lake_ExceptionRaised_jsonCoversionError(client):
```
payneBrandon, line 97: "ConversionError" rather than "CoversionError".

tests/test_main.py
```diff
@@ -56,20 +66,45 @@ def test_rsu_data_lake_bucket(client):
...
-    raw_blob.upload_from_string('{"column":1, "status":"yes"}')
-    lake_bucket = 'rsu_data-lake'
-    lake_bucketOBJ = client().get_bucket(lake_bucket)
-    main.help_data_lake(raw_blob, raw_bucketOBJ, lake_bucketOBJ)
+    json_bytes = b'{"timeReceived": "2020-05-14T11:37:06Z", "year": "2020", "month": "05", "day": "14", "hour": "11", "version": "1.1.0", "type": "bsm"}'
+    raw_bucket = client().get_bucket('rsu_raw-ingest')
+    lake_bucket = client().get_bucket('rsu_data-lake')
+    raw_blob = raw_bucket.blob("test")
+
+    raw_blob.upload_from_string('{"timeReceived": "2020-05-14T11:37:06Z", "year": "2020", "month": "05", "day": "14", "hour": "11", "version": "1.1.0", "type": "bsm"}')
```
payneBrandon, line 90: I don't think we are gaining anything by calling this `upload_from_string` method. We don't need to mock it, nor are we doing so here, just calling as though it were a method.

tests/test_main.py
```diff
@@ -56,20 +66,45 @@ def test_rsu_data_lake_bucket(client):
...
+    raw_blob.download_as_bytes.return_value = json_bytes
 
-    assert True
+    main.verify_and_copy_blobs_to_data_lake([raw_blob], raw_bucket, lake_bucket)
+    raw_bucket.copy_blob.assert_called_with(raw_blob, lake_bucket)
```
payneBrandon, line 94: This test handles the case where the JSON is clean. We'll want another test to show what happens when it is not clean. The easiest way would be to just mock the `is_json_clean` method, then we don't have to pass it real data.

tests/test_main.py
payneBrandon: Actually, I don't see a method called "help_data_lake" and we aren't calling any method from the tested class so I don't think this test needs to be here.

tests/test_main.py
```diff
@@ -56,20 +66,45 @@ def test_rsu_data_lake_bucket(client):
...
+    lake_bucket = client().get_bucket('rsu_data-lake')
+    raw_blob = raw_bucket.blob('test')
+
+    raw_blob.upload_from_string('not a valid json string')
+    raw_blob.download_as_bytes.return_value = invalid_json_bytes
+
+    assert not raw_bucket.copy_blob.called
 
 @mock.patch("google.cloud.pubsub_v1.PublisherClient")
 @mock.patch("google.cloud.storage.Client")
```
payneBrandon, line 110: This would be a great method to inject the `download_and_publish_blobs_to_data_warehouse` method and verify it was called.

tests/test_main.py
```diff
@@ -81,7 +116,30 @@ def test_rsu_data_warehouse_bucket(client, publish_client):
...
+
+    with pytest.raises(exceptions.NotFound):
+        main.rsu_data_warehouse_bucket(publish_client(), client(), topic_path, data_lake_bucket)
+    
+    assert not mockHelpDataWarehouse.called
+
+@mock.patch.object(main, "download_and_publish_blobs_to_data_warehouse")
+@mock.patch("google.cloud.pubsub_v1.PublisherClient")
+@mock.patch("google.cloud.storage.Client")
+def test_rsu_data_warehouse_ExceptionRaised_TopicNotFound(client, publish_client, mockHelpDataWarehouse):
```
payneBrandon, line 136: We're not calling out to our tested class here.

tests/test_main.py
```diff
@@ -92,17 +150,5 @@ def test_help_warehouse(client, publish_client):
...
-
-    data = 'data'
-    context = 'context'
-    main.main(data, context)
-
-    assert True
-
-
+    main.download_and_publish_blobs_to_data_warehouse([lake_blob], publish_client(),topic_path)
+    publish_client().publish.assert_called_with(topic_path, client().get_bucket().blob().download_as_bytes())
```
payneBrandon, line 154: Since we pass in an array of blobs, we should verify that the publish method is called the same number of times as our array count.

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,115 @@
...
+    lake_bucket = 'rsu_data-lake'
+    lake_bucketOBJ = client().get_bucket(lake_bucket)
+
+    raw_to_data_lake.raw_to_data_lake(raw_bucketOBJ, lake_bucketOBJ, raw_blob)
+
+    assert raw_to_data_lake.is_json_clean(raw_blob.download_as_bytes.ndjson.loads) is True
+
+@mock.patch.object(raw_to_data_lake, "is_json_clean", return_value=True)
+@mock.patch("google.cloud.storage.Client", autospec=True)
+def test_raw_to_data_lake_success_blob_copied(client, mockIsJSONCleanFunction):
```
payneBrandon, line 53: Rather than set up each of these buckets and bytes with actual data, you could do something like the following:
```
@mock.patch.object(raw_to_data_lake, "is_json_clean", return_value=True)
@mock.patch("google.cloud.storage.Client", autospec=True)
@mock.patch.object(raw_to_data_lake, "json")
def test_raw_to_data_lake_success_blob_copied(json, client, mockIsJSONCleanFunction):
    raw_bucketOBJ = mock.Mock()
    lake_bucketOBJ = mock.Mock()
    raw_blob = mock.Mock()

    raw_to_data_lake.raw_to_data_lake(raw_bucketOBJ, lake_bucketOBJ, raw_blob)
    raw_bucketOBJ.copy_blob.assert_called_with(raw_blob, lake_bucketOBJ)
```
This ends up being quite a bit cleaner, and faster due to not having to set up real data. We only really want to verify what happens when we pass valid data, so we can mock up the rest of it.

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,115 @@
...
+    lake_bucketOBJ = client().get_bucket(os.environ['data_lake_id'])
+
+    raw_blob.upload_from_string('{"timeReceived": "2020-05-14T11:37:06Z", "year": "2020", "month": "05", "day": "14", "hour": "11", "version": "1.1.0", "type": "bsm"}')
+    raw_blob.download_as_bytes.return_value = json_bytes
+
+    raw_to_data_lake.raw_to_data_lake(raw_bucketOBJ, lake_bucketOBJ, raw_blob) 
+    raw_bucketOBJ.copy_blob.assert_called_with(raw_blob, lake_bucketOBJ)
+
+@mock.patch("google.cloud.storage.Client", autospec=True)
+def test_raw_to_data_lake_Exception_jsonConversionError(client):
```
payneBrandon, line 70: Here we aren't calling anything in the tested class, just setting up variables.

tests/test_raw_to_data_lake.py
```diff
@@ -0,0 +1,115 @@
...
+    # Act
+    with pytest.raises(exceptions.NotFound):
+        raw_to_data_lake.main(event, context)
+
+    # Assert
+    assert not mockRawToDataLakeFunction.called
+
+@mock.patch.object(raw_to_data_lake, "raw_to_data_lake")
+@mock.patch("google.cloud.storage.Client", autospec=True)
+def test_main_ExceptionRaised_BlobNotFound(mockClient, mockRawToDataLakeFunction):
```
payneBrandon, line 104: Again, we're not calling into the tested class.

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,83 @@
+from GCP_cloud_functions.lake_to_data_warehouse import lake_to_data_warehouse
+import mock
+from google.api_core import exceptions
```
payneBrandon, line 3: Several of these imports are not needed, can we remove?

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,83 @@
...
+    l_bucket = client().get_bucket('test_data_lake_bucket')
+    lake_blob = l_bucket.blob('test')
+    lake_blob.upload_from_string('{"timeReceived": "2020-05-14T11:37:06Z", "year": "2020", "month": "05", "day": "14", "hour": "11", "version": "1.1.0", "type": "bsm"}')
+
+    lake_to_data_warehouse.rsu_data_warehouse_bucket(publish_client(), topic, lake_blob)
+    publish_client().publish.assert_called_with(topic, client().get_bucket().blob().download_as_bytes())
+
+@mock.patch.object(lake_to_data_warehouse, "rsu_data_warehouse_bucket")
+@mock.patch("google.cloud.storage.Client")
+def test_main_NotCalled_BlobNotFound(mockClient, mockLakeToDataWarehouseFunction):
```
payneBrandon, line 22: Not calling to our class we are testing

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,83 @@
...
+    context = None
+    l_bucket = mockClient().get_bucket(event['bucket'])
+    l_bucket.get_blob.side_effect = None    # get_blob returns None if blob does not exist
+
+    assert not mockLakeToDataWarehouseFunction.called
+
+@mock.patch.object(lake_to_data_warehouse, "rsu_data_warehouse_bucket")
+@mock.patch("google.cloud.pubsub_v1.PublisherClient")
+@mock.patch("google.cloud.storage.Client")
+def test_main_NotCalled_TopicNotFound(mockClient, mockPublisher, mockLakeToDataWarehouseFunction):
```
payneBrandon, line 40: Not calling to our tested class

tests/test_lake_to_data_warehouse.py
```diff
@@ -0,0 +1,83 @@
...
+
+    topic = mockPublisher.topic_path(os.environ['project_id'], os.environ['data_hub_id'])
+    mockPublisher.topic_path.side_effect = None
+
+    assert not mockLakeToDataWarehouseFunction.called
+
+@mock.patch.object(lake_to_data_warehouse, "rsu_data_warehouse_bucket")
+@mock.patch("google.cloud.pubsub_v1.PublisherClient")
+@mock.patch("google.cloud.storage.Client")
+def test_main_Success(mockClient, mockPublisher, mockLakeToDataWarehouseFunction):
```
payneBrandon, line 56: We're mocking up quite a few objects in here that aren't being used. If we look at the `main` function to be tested, we use publisherClient, storage.Client, and event objects. Those should be the only ones we need to mock. So you could bring this test method down to something like the following:
```
@mock.patch.object(lake_to_data_warehouse, "rsu_data_warehouse_bucket")
@mock.patch("google.cloud.pubsub_v1.PublisherClient")
@mock.patch("google.cloud.storage.Client")
def test_main_Success(mockClient, mockPublisher, mockLakeToDataWarehouseFunction):
    event = mock.MagicMock()
    context = mock.MagicMock()    
    os.environ['project_id'] = 'project_id'
    os.environ['data_hub_id'] = 'hub_id'
    lake_to_data_warehouse.main(event, context)

    mockPublisher().topic_path.assert_called_with(os.environ['project_id'], os.environ['data_hub_id'])
    mockClient().get_bucket.assert_called_with(event['bucket'])
    mockClient().get_bucket(event['bucket']).get_blob.assert_called_with(event['name'])
    assert mockLakeToDataWarehouseFunction.called
```

data_manager/source_code/main.py
dhivahari-vivek-cdot: Oh, sorry. I think it's just pytest that I need to remove? Done.

tests/test_main.py
```diff
@@ -92,17 +150,5 @@ def test_help_warehouse(client, publish_client):
...
-def test_main():
-
-    data = 'data'
-    context = 'context'
-    main.main(data, context)
-
-    assert True
-
-
+    main.download_and_publish_blobs_to_data_warehouse([lake_blob], publish_client(),topic_path)
```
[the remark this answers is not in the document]
dhivahari-vivek-cdot: Done.

tests/test_main.py
dhivahari-vivek-cdot: I think I'm actually removing this because the actual topic is created in main, not in this function, so the error would be thrown in main and wouldn't even get to this point.

tests/test_main.py
dhivahari-vivek-cdot: I see your example below. Let me go back and try that.

tests/test_main.py
dhivahari-vivek-cdot: You are right! Removed.

tests/test_main.py
dhivahari-vivek-cdot: Fixed typo. And aren't I checking that copy_blob doesn't get called if the JSON cleanliness test doesn't pass? I modified the test here so that we do call the method, lmk if we still want to remove?

tests/test_main.py
dhivahari-vivek-cdot: Done.

tests/test_main.py
dhivahari-vivek-cdot: Done.

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: Quick question, actually: do we still want to test that the blob was not found here? Here, the event (which was a new blob added to the `rsu_raw_ingest` bucket) is what triggers this function, and in the code, we're passing in that `event['name']` to the `get_blob` function. How would this function ever result in a blob not being found?

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: Using your proposed faster method. Looks I keep forgetting the "act" part of the testing process. Done!

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: I see what you mean now. My apologies, it seems I misunderstood your explanation before. This does seem a lot easier. Thanks! Done.

A question for me, though: when we make raw_blob, how are we connecting that to the `mockIsJSONCleanFunction` argument we passed. In fact, how are we making those buckets without referencing the `client()` in, say, `client().get_bucket`. By calling `mock.Mock()` for all of those, the code just knows to attribute it to those specific patches?

tests/test_lake_to_data_warehouse.py
dhivahari-vivek-cdot: Done.

tests/test_lake_to_data_warehouse.py
dhivahari-vivek-cdot: Same question as above w.r.t. it passing the `name` and `bucket` attributes of the `event`.

tests/test_main.py
dhivahari-vivek-cdot: Nevermind! The `topic_path` function in main returns the `str` of the path, and that same string is passed into the `rsu_data_warehouse` function. It looks like the actual error is thrown in the helper function `download_and_publish_blobs_to_data_warehouse()`. Give me a second to figure out which method is actually throwing it!

tests/test_lake_to_data_warehouse.py
dhivahari-vivek-cdot: Same comment with the topic exception above. Let me see where it's actually being thrown.

tests/test_lake_to_data_warehouse.py
dhivahari-vivek-cdot: I see. So I was doing unnecessary work by creating all these variables. Thank you for showing me this. Fixed.

tests/test_main.py
```diff
@@ -1,15 +1,10 @@
-import sys
-sys.path.append("..")
 from data_manager.source_code import main
 import mock
 from unittest.mock import patch, mock_open
```
payneBrandon, line 3: These aren't used, can we remove?

tests/test_main.py
```diff
@@ -1,15 +1,10 @@
...
 from unittest import TestCase
```
payneBrandon, line 4: Same here, `TestCase` is never used.

README.md
payneBrandon: @dhivahari-vivek-cdot we're still missing these environment variables in the README here. Can we get those added?

payneBrandon:
> It looks like we've removed the config usage, can we remove the import in `main.py` and remove the file? As a side note here, there seems to be other unused imports in that `main.py` file we can remove as well.

@dhivahari-vivek-cdot ?

tests/test_main.py
payneBrandon: We still have several unused imports:
```
from unittest.mock import patch, mock_open
from unittest import TestCase
```
don't seem to be used

data_manager/source_code/main.py
payneBrandon: Looks like json is also unused. We may  need to look at your environment to make sure you get this same functionality (looks like it comes through with the [Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) extension) but typically you'll see something like this:
image

Notice how in the `import json` statement "json" is greyed out. This means it's not being used, and we can remove it. Hovering over the import statement confirms this: 
image

tests/test_main.py
payneBrandon: Now that we're calling the method, the unit test makes sense. Without that call, we do all the work and no actual test.

tests/test_main.py
payneBrandon: @dhivahari-vivek-cdot any luck here?

tests/test_raw_to_data_lake.py
payneBrandon: In this case, we don't have to have the raw_blob connected to anything. The method we are testing takes in three parameters, all of which we mock up using `mock.Mock()` which essentially just creates a random mock object. We don't need the functionality the real library provides though so we can get away without injecting a fully mocked object. 

We further mock the `is_json_clean` method and say "whatever you send me, I'll just return true". So now we have mocked libraries and mocked method, and we can test the logic in this individual method. You could turn this method around and patch in all the full mocks rather than `mock.Mock` if you wanted though.

tests/test_raw_to_data_lake.py
payneBrandon: We seem to have lost this exception test case.

tests/test_raw_to_data_lake.py
payneBrandon: Ideally it wouldn't happen, but what if the method was fired off twice, or someone down the line manually calls it?

tests/test_lake_to_data_warehouse.py
payneBrandon: Still including unused imports. `exceptions` & `pytest` are not used.

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: Thank you! Done.

tests/test_main.py
dhivahari-vivek-cdot: Done.

data_manager/source_code/main.py
dhivahari-vivek-cdot: Done.

tests/test_main.py
dhivahari-vivek-cdot: Done.

tests/test_lake_to_data_warehouse.py
dhivahari-vivek-cdot: Done.

tests/test_main.py
dhivahari-vivek-cdot: Done.

tests/test_main.py
dhivahari-vivek-cdot: Done.

tests/test_raw_to_data_lake.py
dhivahari-vivek-cdot: Done?

PR #20, opened 2021-04-22
RSU MsgFwd SNMP Configuration Automation

drewjj:
This includes:
- The script for the automation of configuring message forwarding over remote SNMP
- SNMP MIB profiles for configuring a new VM for externally running SNMP commands to the RSU
- Guide on how to use the script and setting up SNMP
- Unit tests for code coverage and branch coverage *SNMP tests are mainly for code coverage


automation/README.md
```diff
@@ -0,0 +1,50 @@
+# SNMP RSU Configuration Automation for Message Forwarding
+This directory contains a script that is used for configuring a large number of CDOT Kapsch RSUs autonomously by providing it a list of RSU IPs and some other arguments.
+
+## What the script does
+The current script can be run manually from a command line shell. It can be ran from internally in an RSU or externally from an RSU. This was developed with running it externally in mind. It has been written in a way that can easily be transformed to be invoked in other methods, such as a REST API, but these are not supported as is.
```
payneBrandon, line 5: Are you meaning here that the script can be ran from within an RSU or from the GCP network?

automation/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+- The destination IP. Example: `10.0.1.5`
+- The UDP port the RSU should forward the packets to at the destination IP. Example: `46800`
+- The RSU index the SNMP configuration should be written to on the RSU. Example: `20`
+
+These will be provided to the script in the form of arguments.
+
+Check out the example shell command to run the script in a terminal:
+```
+python3 configrsu_msgfwd.py /home/user/RSU_Management/tests/test_files/snmp_test.csv 10.0.1.5 46800 20
```
payneBrandon, line 29: Is the python script expecting a text file or csv? Above we have a text file that appears to be separated by newline characters, but here you're loading a csv file.

automation/README.md
drewjj: Yes, it can

automation/README.md
drewjj: A csv. Currently all that is needed is a single column csv

payneBrandon: Let's update that line 14 to specify csv rather than text file then

automation/README.md
payneBrandon: Can we reword to make that more clear then? This text is saying it ca be ran internally from an RSU or externally from an RSU which isn't very clear.

automation/README.md
```diff
@@ -0,0 +1,50 @@
...
+- A text file containing a list of RSU IP addresses this can look like the following:
+```
+10.0.0.1
+10.0.0.2
+10.0.0.3
+10.0.0.4
+```
+- The destination IP. Example: `10.0.1.5`
+- The UDP port the RSU should forward the packets to at the destination IP. Example: `46800`
+- The RSU index the SNMP configuration should be written to on the RSU. Example: `20`
```
payneBrandon, line 23: Is there a standard SNMP config index we should default to? I know with WYDOT we allowed indexes 1-99 to be used by TIMs and reserved the first index for config values. We'll likely want to do something similar so we know what indexes we can use for other purposes.

automation/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+These will be provided to the script in the form of arguments.
+
+Check out the example shell command to run the script in a terminal:
+```
+python3 configrsu_msgfwd.py /home/user/RSU_Management/tests/test_files/snmp_test.csv 10.0.1.5 46800 20
+```
+## Setting up SNMP on external machine
```
payneBrandon, line 31: Might be for next iteration, but what about creating a docker image we could spin up with all this already installed? Then we'd just pull the image and be rolling.

automation/README.md
```diff
@@ -0,0 +1,50 @@
...
+```
+sudo apt update
+sudo apt install snmpd
+sudo apt install snmp
+```
+Now create or copy the correct MIB profiles in the `/usr/share/snmp/mibs` directory
+
+List of required MIB profiles:
+```
+IPV6-TC.txt
```
payneBrandon, line 46: Can we include a brief description of what each of these profiles is doing?

automation/README.md
drewjj: From what Tony and I have heard from CDOT and Kapsch there is not. The documentation says to pick a number above 800 but Ashley told us to just use 1 at one point. It seems like it is up to us to make a standard.

automation/README.md
drewjj: This could definitely work. I would have to rethink how this is executed to be ran properly on a docker image, though. Definitely worth keeping in mind for the next iteration.

automation/README.md
drewjj: I can try, I honestly am not sure myself.

drewjj: @payneBrandon gave it a shot

automation/README.md
payneBrandon: Above 800? How many indexes are on these Kapsch units?

drewjj: Apparently a lot! They no problem with me putting a message forward config on index 1050 when I shared with them my config commands

automation/configrsu_msgfwd.py
```diff
@@ -0,0 +1,59 @@
+import argparse
```
payneBrandon, line 1: Doesn't look like we're using this import - can we remove?

automation/configrsu_msgfwd.py
```diff
@@ -0,0 +1,59 @@
+import argparse
+import csv
+import os
+import sys
+
+os.environ['SNMP_USERNAME'] = 'authOnlyUser'
```
payneBrandon, line 6: Let's remove these from getting set in this file and mention that they are required in the README. We don't want any sensitive data in this repo.

automation/configrsu_msgfwd.py
```diff
@@ -0,0 +1,59 @@
...
+  # Put RSU in run mode
+  set_rsu_status(rsu_ip, operate=True)
+  
+  os.system('snmpwalk -v 3 -u {auth} {rsuip} 1.0.15628.4.1 | grep 4.1.7'.format(auth=snmp_authstring, rsuip=rsu_ip))
+
+def main(rsu_csv, dest_ip, udp_port, rsu_index):
+  for row in rsu_csv: 
+    config_msgfwd(row[0], str(dest_ip), str(udp_port), str(rsu_index))
+
+if __name__ == '__main__':
```
payneBrandon, line 55: Just for posterity, can we add a comment in here stating what we're expecting or arguments? I know we already have it in the README but it's handy to see it in the code as well.

payneBrandon: Actually, I see you have a good naming convention in the main function that this is calling so that's explanatory enough. Closing this comment.

automation/configrsu_msgfwd.py
```diff
@@ -0,0 +1,59 @@
...
+    os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuMode.0 i 4'.format(auth=snmp_authstring, rsuip=rsu_ip))
+  else:
+    os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuMode.0 i 2'.format(auth=snmp_authstring, rsuip=rsu_ip))
+
+def config_msgfwd(rsu_ip, dest_ip, udp_port, rsu_index):
+  # Put RSU in standby
+  set_rsu_status(rsu_ip, operate=False)
+  
+  # Create a little endian hex version of destIP
+  hex_dest_ip = ip_to_hex(dest_ip, little_endian=True)
```
payneBrandon, line 33: Is this little/big endian something that we should add as a parameter to the script? I see the `ip_to_hex` method is set up to handle a boolean value for it, so it might make sense to abstract that to a parameter.

automation/configrsu_msgfwd.py
```diff
@@ -0,0 +1,59 @@
...
+  
+  # Perform configurations
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdPsid.{index} s " "'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdDestIpAddr.{index} x {destip}'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index, destip=hex_dest_ip))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdDestPort.{index} i {port}'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index, port=udp_port))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdProtocol.{index} i 2'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdRssi.{index} i -100'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdMsgInterval.{index} i 1'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdDeliveryStart.{index} x 0C1F07B21100'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
```
payneBrandon, line 42: What are these hex values we're setting for the delivery start/stop values?

automation/configrsu_msgfwd.py
```diff
@@ -0,0 +1,59 @@
...
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdDeliveryStop.{index} x 0C1F07F41100'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdEnable.{index} i 1'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+
+  # Put RSU in run mode
+  set_rsu_status(rsu_ip, operate=True)
+  
+  os.system('snmpwalk -v 3 -u {auth} {rsuip} 1.0.15628.4.1 | grep 4.1.7'.format(auth=snmp_authstring, rsuip=rsu_ip))
```
payneBrandon, line 49: What are we printing out here? 4.1 looks to be the RSU Spec, but I'm not sure what the `.7` does here.

automation/configrsu_msgfwd.py
```diff
@@ -0,0 +1,59 @@
...
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdDestPort.{index} i {port}'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index, port=udp_port))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdProtocol.{index} i 2'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdRssi.{index} i -100'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdMsgInterval.{index} i 1'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdDeliveryStart.{index} x 0C1F07B21100'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdDeliveryStop.{index} x 0C1F07F41100'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
+  os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuDsrcFwdEnable.{index} i 1'.format(auth=snmp_authstring, rsuip=rsu_ip, index=rsu_index))
```
payneBrandon, line 44: I'm not super familiar with this `os.system` command, but is it possible that the snmpset command would fail and cause an exception which would leave our RSU in a downed state?

automation/configrsu_msgfwd.py
drewjj: Yes, that can be removed

automation/configrsu_msgfwd.py
drewjj: Right, will do

automation/configrsu_msgfwd.py
drewjj: that is probably something that should be added, yeah

automation/configrsu_msgfwd.py
drewjj: They are datetimes in hex. One for the start and one for the end. They translate to Dec 31 1970 and Dec 31 2036. This is just what values Ashley's example commands were using so I left them as is. This would obviously need to be updated come 2037 haha.

automation/configrsu_msgfwd.py
drewjj: This prints out the configuration for message forwarding specifically. I imagine that is what the .7 refers to as well.

automation/configrsu_msgfwd.py
drewjj: Possible I suppose. I can't imagine it failing after the first one succeeded at setting to standby but if there were network issues I cannot imagine we would be able to revert it regardless.

It wouldn't throw an exception, though. It would just fail in the output.

automation/configrsu_msgfwd.py
payneBrandon: Ok that makes sense. Can we add a comment saying what dates they correspond with?

drewjj:
Changes should be good to go

automation/README.md
payneBrandon: Let's pick one as our standard and add some comments in here about that. We could even make the index an optional parameter and default it to our standard value. That way we know what index to look at when querying in the future.

tests/snmptest.py
```diff
@@ -7,14 +7,17 @@
```
payneBrandon, line 7: I'm running into errors running the test on this file. The pytest doesn't seem to know what the `configrsu_msgfwd` is. If you change the import to `from automation import configrsu_msgfwd` then it imports properly, but I'm still seeing issues with finding the file in the `test_main` method. If I change the file fetch to the following it runs:
```
real_path = os.path.realpath(__file__)
dir_path = os.path.dirname(real_path)
file = os.path.join(dir_path, 'test_files', 'snmp_test.csv')
```

tests/snmptest.py
```diff
@@ -0,0 +1,59 @@
+import sys
+sys.path.append("../automation/")
+import csv
+import os
+import pytest
```
payneBrandon, line 5: `pytest` and `TestCase` don't appear to be used, can we remove?

tests/snmptest.py
drewjj: Hmm, strange. It runs just fine on my machine... I wonder why it needs the changes. Well I can implement what works on your machine and test it on mine later this afternoon.

tests/snmptest.py
drewjj: Yeah, I was just making sure to import what the existing test was utilizing since it was not clear to me what unit test suite we were using.

tests/snmptest.py
```diff
@@ -0,0 +1,59 @@
...
+  hex = configrsu_msgfwd.ip_to_hex(ip, 1)
+  assert hex == '08080808000000000000000000000000'
+
+def test_rsu_status_off():
+  rsu_ip = '0.0.0.0'
+  
+  # Since there is no snmp server, simply running the shell command successfully should suffice
+  try:
+    configrsu_msgfwd.set_rsu_status(rsu_ip, False)
+    assert True
```
payneBrandon, line 29: Rather than `assert True/false` in these unit tests, can we look into what is called and verify that? For instance, this function we should mock the os injection along with the env variables and verify that we turned operate off. IE ensure that the following was called:
```
os.system('snmpset -v 3 {auth} {rsuip} RSU-MIB:rsuMode.0 i 2'.format(auth=snmp_authstring, rsuip=rsu_ip))
```

drewjj: This is reasonable, I would need to look into how to do this.

payneBrandon: We're doing some of this in the Work_Zone repo, https://github.com/CDOT-CV/Work_Zone/blob/main/test/test_main.py has an example of patching in the os variables as a starting point.

drewjj: this one is definitely resolved, it no longer asserts True

automation/README.md
drewjj: This comment is outdated because now we configure the index since RSUs can forward to multiple endpoints for the same message type.

automation/README.md
```diff
@@ -0,0 +1,56 @@
+# SNMP RSU Configuration Automation for Message Forwarding
+This directory contains a script that is used for configuring a large number of CDOT Kapsch RSUs autonomously by providing it a list of RSU IPs and some other arguments.
```
payneBrandon, line 2: If this is only for the Kapsch RSUs, should we rename the `configrsu_msgfwd.py` script to be `configKapschrsu_msgfwd.py` to emphasize?

drewjj: That is a good point. I can do that.

drewjj: Updated this.

payneBrandon (approved):
This looks good to me!

PR #22, opened 2021-09-21
BSM Count View

drewjj:
Includes:
- BSM count side panel
- Sortable table with BSM data
- Loading circle animation
- RSU query count cloud function update
- API Gateway update to adjust for changes to the cloud function


cdot-web-app/src/App.js
```diff
@@ -18,20 +26,81 @@ class App extends Component {
 
   async getRsuData() {
     try {
-      let response = await axios.get('https://rsu-manager-apigateway-5xm3e3o5.uc.gateway.dev/rsuinfo');
-      this.setState({'rsuData': response.data.rsuList});
+      let rsuInfoData = await axios.get('https://rsu-manager-apigateway-5xm3e3o5.uc.gateway.dev/rsuinfo');
```
payneBrandon, line 29: We can leave it hard-coded for this PR, but let's look at pulling this address out to a config file for each environment for our next steps.

cdot-web-app/src/components/Header.js
```diff
@@ -20,9 +20,10 @@ const style = {
...
-  padding: '10px',
-  position: 'relative',
-  zIndex: '100'
+  position: 'absolute',
+  zIndex: '100',
+  width: '100%',
+  height: '100px'
 }
 
 const gridStyle = {
```
payneBrandon, line 29: Should this be "alignment"?

cdot-web-app/src/components/Header.js
```diff
@@ -31,7 +32,7 @@ const gridStyle = {
 
 const logoStyle = {
   padding: '5px',
-  marginLeft: '20px',
+  height: '0%',
   alignment: 'center'
 }
```
payneBrandon, line 38: Is this 0% height doing anything?

cdot-web-app/src/App.js
```diff
@@ -18,20 +26,81 @@ class App extends Component {
...
+      let rsuCountsData = await axios.get('https://rsu-manager-apigateway-5xm3e3o5.uc.gateway.dev/rsucounts');
+      this.setState({'rsuData': rsuInfoData.data.rsuList});
+      this.setState({'rsuCounts': rsuCountsData.data});
+
+      var workingList = [];
+      Object.entries(rsuCountsData.data).map(([key, value]) => {
```
payneBrandon, line 35: Rather than setting an external list, the map function should just return values:

```
var workingList = Object.entries(rsuCountsData.data).map(([key, value]) => {
        console.log(key);
        return {
          "key": key,
          "rsu": key,
          "road": value.road,
          "count": value.count
        };
      });
```
Note that I did add a "key" field here to get around an oddity with React wanting unique keys.

cdot-web-app/src/components/Header.js
drewjj: Nope, that was an artifact of something I tried. I will remove that.

cdot-web-app/src/components/Header.js
drewjj: I am unsure on what you are referring to here. The CSS elements seem to work fine and I referenced some online resources for making elements display over other elements to make the header and the mapbox map play nicely and they said to use absolute.

payneBrandon: It wouldn't let me add a comment in the right spot I guess. Just below this is "gridStyle" style with a css property "alightnment" set to center. It should be "alignment".

cdot-web-app/src/App.js
drewjj: I think this has been addressed

cdot-web-app/src/components/Header.js
drewjj: Addressed

cdot-web-app/src/components/Header.js
drewjj: I see. I must have been not looking haha. I fixed it.

cdot-web-app/src/App.js
drewjj: Okay, I tried fixing it but for some reason it wasn't properly accessing the environment variables. I didn't want to fiddle with it for too long so I will just make sure this is fixed with my update to this repo this sprint.

payneBrandon (approved):
Looks good!

PR #24, opened 2021-09-23
Database/rsu insert ping

payneBrandon:
New python function to insert historic ping data. Includes `ping-insert-test.py` and `simpleRequest.py` for easier testing.

Entry method expects a JSON object in the request body. An example of this can be found in the `ping-insert-test.py` file, "testJson" object.


drewjj (approved):
Looks good, looks like we will need the API Gateway to reference this now. I know you are working on that.

PR #25, opened 2021-10-07
Feat/function deploy

payneBrandon:
Updates `azure-pipelines.yml` file to split out functions into separate zip files and upload to storage bucket for function use


drewjj:
Looks good, so this is getting automatically consumed by the cloud function?

payneBrandon:
> Looks good, so this is getting automatically consumed by the cloud function?

I'll need to run some tests with an already deployed function app to see how it behaves with just pushing a new zip file. For now, this deploys appropriate zipped files for an initial deployment. I may end up adding a "deploy function" portion to the script as well.

drewjj (approved):
Okay, sounds good.

PR #28, opened 2021-10-27
Static RSU user help page

ravishersingh2:
- Added a static user help page.
- Converted function based component to class based component for map.js


payneBrandon:
> We need a few formatting updates in here. First, the banner is covering the text:
> 
> image
> 
> Second, each line in there is set up as a bullet point. This is not needed and reads a bit funny.

@ravishersingh2 This still seems to be hidden behind the banner.

cdot-web-app/src/components/documentation.js
```diff
@@ -0,0 +1,46 @@
...
+     
+        
+      
+
+
+
+<li>This application shows the location of RSU units installed by CDOT on various roads of Colorado where RSU units are responsible for collecting the data.</li>
+
+<li>Each RSU unit is shown in red,green  and yellow color points on road and on clicking the point it will give you following about that unit(see fig below).</li>
+<li>4.	If a point is red, it means that unit is offline.</li>
```
payneBrandon, line 25: The numbering in here is random, let's clean that up a bit.

payneBrandon:
@ravishersingh2 I've uploaded a few changes with wording and formatting, as well as a little bit of reordering in the descriptions. Can we keep cleaning that up a bit in the same fashion?

payneBrandon:
@ravishersingh2 getting closer! I've pushed up a few changes for formatting and naming convention updates, please pull before making additional changes. Now that we have the header issue fixed, let's remove the scrollbar on the right - we have a small page and the scrollbar is not needed here.

cdot-web-app/src/components/Documentation.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
```
payneBrandon, line 1: I'd made a few updates to this in a (previous commit)[https://github.com/CDOT-CV/RSU_Management/pull/28/commits/8cb814cf0713ac39688abd202b5c00ab824cc94a] that you seem to have overwritten. Take a look at those and update here as needed.

cdot-web-app/src/components/Header.js
```diff
@@ -1,7 +1,9 @@
 import React, { Component } from 'react';
 import logo from '../images/cdot_logo.png';
-
+import './css/Menu.css'
 import Grid from '@material-ui/core/Grid';
+import './Menu.js';
+import Documentation from './Documentation';
```
payneBrandon, line 6: Doesn't seem to be used, can we remove?

cdot-web-app/src/index.html
```diff
@@ -0,0 +1,163 @@
+<!-- //key token
```
payneBrandon, line 1: I don't think we need this file. Everything is just commented out

cdot-web-app/src/components/Documentation.js
payneBrandon: @ravishersingh2, did you look over that change? This documentation outline is still cluttered and difficult to read.

payneBrandon (changes requested):
The main map page no has a scroll bar that moves the user off the map. This should only be applied to the BSM Counts item on the right.

Also, for some reason I'm no longer seeing RSUs along the highways. IE the red, yellow, green dots are missing on the map.

PR #29, opened 2021-11-02
Update RSU counts query to the new table schema

drewjj:
Includes the updates to the cloud function for rsu counts so that the new table schema can be supported.

Still grabs yesterday's counts (UTC)


drewjj:
One other thing to note, this is hard coded for BSM counts only. This was made to be able to support other message types in the future when we want to.

GCP_cloud_functions/rsu-manager/rsu-querycounts/main.py
```diff
@@ -1,26 +1,29 @@
...
+            f"GROUP BY RSU, Road " \
             f"ORDER BY Road, RSU"
     query_job = client.query(query)
 
     result = {}
     for row in query_job:
       item = {}
       item["road"] = row["Road"]
-      item["count"] = row["BsmCount"]
+      item["count"] = row["Count"]
```
payneBrandon, line 26: What do you think about adding in the 'yesterday' value here as a 'Date' field? That way we know for sure what date is being applied when we hit the service?

drewjj: I added this to the function. I can add this to the RSU Manager to display too but maybe we should wait to add it to the RSU Manager so we can decide where the most useful place to put this value might be.

payneBrandon: Sounds reasonable to me!

payneBrandon (approved):

PR #30, opened 2021-11-05
RSU Manager RSU Online Status

drewjj:
This pull request includes all of the changes added to support displaying RSU online status based off the real Zabbix ping data stored in the PostgreSQL database.

Includes:
- New cloud function that queries for last 15 minutes of ping data to base the current online status of an RSU as either "online", "unstable", "offline" and queries for the last timestamp each RSU was online
- Azure DevOps pipeline changes to the YML file to add the new cloud function to the build pipeline
- Changes to the cdot webapp to change the RSU icon color based off of online status and to also display the last time the RSU was online


GCP_cloud_functions/rsu-manager/rsu-online-status/main.py
```diff
@@ -0,0 +1,170 @@
...
+  num_records = 3
+  result = {}
+  # query ping data
+  ping_result = get_ping_data(num_records)
+
+  # calculate online status
+  for key, value in ping_result.items():
+    result[key] = {}
+    result[key]['current_status'] = None
+    for x in range(num_records):
```
payneBrandon, line 133: I think we'll want to look at this a bit closer, here we're saying that as long as the latest ping is online then we display "online", ignoring any previous failed pings. I think if we see any failed pings in our list we should mark it as unstable.

drewjj: I agree. I based this off what Tony was saying about this value being a clear indication of whether you could reach the RSU at the very moment you were looking at the data so if the last ping was successful it would be online since it is accessible. I definitely see it being reasonable to mark it as unstable but then it may be confusing why you can communicate with some unstable RSUs and not others.

azure-pipelines.yml
```diff
@@ -55,7 +65,9 @@ steps:
       gsutil cp $(Build.ArtifactStagingDirectory)/rsuInfo.zip gs://cf-deployments-$(project_name)
       gsutil cp $(Build.ArtifactStagingDirectory)/rsuQueryCounts.zip gs://cf-deployments-$(project_name)
       gsutil cp $(Build.ArtifactStagingDirectory)/rsuPingInsert.zip gs://cf-deployments-$(project_name)
+      gsutil cp $(Build.ArtifactStagingDirectory)/rsuOnlineStatus.zip gs://cf-deployments-$(project_name)
```
payneBrandon, line 68: Is this the latest version of the azure-pipelines.yml file? We shouldn't be downloading the secure file, copying the zips, or deploying the functions in this build pipeline anymore.

drewjj: Maybe not?

drewjj: I didn't realize the branch was behind, I pulled the updates and this has been fixed. I will push the changes once I am done with the time based query.

drewjj: Resolved

GCP_cloud_functions/rsu-manager/rsu-online-status/main.py
payneBrandon: Let's push this one to the next sprint then to address. I'd think that would be the definition of unstable, that it may or may not respond because we've been seeing some issues. On the flip side if we mark something as online and they can't talk to it with their next attempt because it has been pinging on and off, that'd cause confusion as well.

payneBrandon (approved):

PR #31, opened 2021-11-15
Added a toggle button to show/hide sidebar

akhtara838:
I added in a button and a toggleButton function so that when the user clicks on the button, the sidebar is hidden and if they click the button when the sidebar is already hidden, it will become visible.


payneBrandon (changes requested):
This is looking good! Show/hide functionality works. Let's try to get the Menu to be over the map though, so when we hide it the map is visible underneath. That way the map is using the maximum real estate on the page.

akhtara838:
I just pushed my changes onto my branch. I changed the map width size from "calc(100% - 350px)'' to "100%", so we're using the entire space and placed the toggle button on the header so that it's always visible to the user.

payneBrandon (dismissed):
@akhtara838 I tweaked the show/hide logic to be more react-based rather than manipulating classes from the DOM. Take a look and let me know if you have any questions with it.

drewjj (approved):
Looks good, just resolve the one conflict remaining

akhtara838:
I fixed all the conflicts. Once it's reviewed, I can merge it in

payneBrandon (approved):
Looks good to me!

PR #32, opened 2021-11-22
Webapp pipeline

payneBrandon:
Adding build pipeline for mapping web app. Includes new yaml file describing pipeline as well as various supporting changes.


cdot-web-app/.env.test
```diff
@@ -0,0 +1,3 @@
+BUILD_PATH='./build-test'
+#REACT_APP_MAPBOX_TOKEN=""
+REACT_APP_GATEWAY_BASE_URL="https://rsu-manager-apigateway-5xm3e3o5.uc.gateway.dev"
```
drewjj, line 3: I assume you are updating these base URLs to be different in DevOps?

cdot-web-app/src/EnvironmentVars.js
```diff
@@ -0,0 +1,9 @@
+class EnvironmentVars {
```
drewjj, line 1: This is cool

drewjj (approved):

cdot-web-app/.env.test
payneBrandon: Good call, I've tweaked this to inject through the build pipeline and exclude the base URL here from source control. This way we won't have those URLs exposed on the public repo.

drewjj (approved):

PR #33, opened 2021-11-30
RSU REST API

drewjj:
This includes:
- New RSU REST API docker image
- Python REST API script that validates arguments before running scripts
- Snmpwalk script to be accessed through the REST API

Not included:
- I forgot we have some unit tests in this repository. I can add unit tests for these scripts going forward if that is wanted
- Snmpset support is not supported in the REST endpoint. The original goal was to get the current SNMP configurations visible so I focused on snmpwalk. I have snmpset partially done and will add it to this pull request if it works in a few hours.


drewjj:
Here is example output from the snmpwalk for 172.16.28.200

```
{
  'snmpwalk': {
    '1': {
      'Message Type': 'BSM',
      'IP': '10.235.1.3',
      'Port': 46800,
      'Protocol': 'UDP',
      'RSSI': -100,
      'Frequency': 1,
      'Start DateTime': '1970-12-31 17:00',
      'End DateTime': '2036-12-31 17:00',
      'Forwarding': 'On',
      'Config Active': 'Enabled'
    },
    '2': {
      'Message Type': 'BSM',
      'IP': '10.14.137.229',
      'Port': 8087,
      'Protocol': 'UDP',
      'RSSI': -100,
      'Frequency': 1,
      'Start DateTime': '1970-12-31 17:00',
      'End DateTime': '2036-12-31 17:00',
      'Forwarding': 'On',
      'Config Active': 'Enabled'
    },
    '3': {
      'Message Type': 'BSM',
      'IP': '10.235.1.36',
      'Port': 46800,
      'Protocol': 'UDP',
      'RSSI': -100,
      'Frequency': 1,
      'Start DateTime': '1970-12-31 17:00',
      'End DateTime': '2036-12-31 17:00',
      'Forwarding': 'On',
      'Config Active': 'Enabled'
    },
    '4': {
      'Message Type': 'BSM',
      'IP': '10.235.1.66',
      'Port': 46800,
      'Protocol': 'UDP',
      'RSSI': -100,
      'Frequency': 1,
      'Start DateTime': '1970-12-31 17:00',
      'End DateTime': '2036-12-31 17:00',
      'Forwarding': 'On',
      'Config Active': 'Enabled'
    }
  }
}
```

payneBrandon:
Let's add in some tests to show what the snmpwalk.py script is doing, and how it will handle failed walk attempts, etc.

automation/snmpwalk.py
```diff
@@ -0,0 +1,96 @@
...
+	'iso.0.15628.4.1.7.1.5': ('Protocol', protocol),
+	'iso.0.15628.4.1.7.1.6': ('RSSI', toint),
+	'iso.0.15628.4.1.7.1.7': ('Frequency', toint),
+	'iso.0.15628.4.1.7.1.8': ('Start DateTime', startend),
+	'iso.0.15628.4.1.7.1.9': ('End DateTime', startend),
+	'iso.0.15628.4.1.7.1.10': ('Forwarding', fwdon),
+	'iso.0.15628.4.1.7.1.11': ('Config Active', active)
+}
+
+def get(rsuip, road):
```
payneBrandon, line 71: Can we add some comments in here explaining what we expect to see in each part of the process? For instance after we run the snmpwalk command, what are we expecting to be returned? What format should the snmp_config returned object be in?

automation/snmpwalk.py
```diff
@@ -0,0 +1,96 @@
...
+	output = ''
+	try:
+		output = subprocess.check_output(cmd, shell=True)
+		output = output.decode("utf-8").split('\n')[:-1]
+	except Exception as e:
+		print(f'SNMP failed to run: {e}')
+		return {}
+	
+	# Placeholder for possible other failed scenarios
+	if len(output) < 10:
```
payneBrandon, line 83: Would we get an output up to 10 that would be failed? Let's explain this a bit with an example, or explanation of why up to 10.

payneBrandon (changes requested):
Looking good, just a couple things to note. Let's also restructure this a bit, maybe adding the txt files under a "resources" directory, code files under "src" or similar. We could also rename the "automation" directory to be more specific like "snmp_automation" so we know what to expect in here.

automation/snmpwalk.py
drewjj: I may have addressed this

automation/snmpwalk.py
drewjj: Addressed with comment in the script

drewjj:
Added a new README and re-organize the automation directory

rsu_automation/README.md
```diff
@@ -0,0 +1,59 @@
+# <b>RSU Automation</b>
+## <b>Purpose</b>
+The RSU Automation docker image allows for a front end application to interface with RSUs over a REST API to read and configure message forwarding settings over SNMP. Future features will be added to this project to support SCMS certification status checking, general RSU maintenance, and J2735 message deployments (such as MAP and TIM) over the REST API.
```
payneBrandon, line 3: Just a note, the "J2735 message deployments" at the end here will likely go through the ODE since it has the ability to encode and sign those message sets. Or are you thinking of a test message that is already encoded/signed and we would expand this library to push those out?

rsu_automation/README.md
```diff
@@ -0,0 +1,59 @@
+# <b>RSU Automation</b>
+## <b>Purpose</b>
+The RSU Automation docker image allows for a front end application to interface with RSUs over a REST API to read and configure message forwarding settings over SNMP. Future features will be added to this project to support SCMS certification status checking, general RSU maintenance, and J2735 message deployments (such as MAP and TIM) over the REST API.
+
+## <b>Current Features</b>
+- REST API utilizing the Python library Flask
+- Message forwarding configurations over SNMP *(Supporting RSUs utilizing SNMP RSU 4.1 spec for message forwarding configuration)*
```
payneBrandon, line 7: Can we link to a location the SNMP RSU 4.1 spec lives for users to look up? I'm not sure if the spec is open source/available or not but if not we can just link to where they can purchase it.

rsu_automation/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM gcr.io/google-appengine/python
```
payneBrandon, line 1: Well that's a handy starting image!

rsu_automation/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN pip3 install flask
```
payneBrandon, line 3: For these pip requirements, would it make sense to add them to a `requirements.txt`, copy over that file and run `pip install -r ./requirements.txt`? That way we wouldn't have to change the Dockerfile in the future and for users hitting it locally they'd know what packages to install.

rsu_automation/README.md
drewjj: Wasn't really certain what features we were implementing through here to be honest. Although the ODE doesn't actually push those out over SNMP to RSUs does it?

rsu_automation/README.md
drewjj: I have no idea. I will ask Tony.

rsu_automation/Dockerfile
drewjj: yeah, we could do that

rsu_automation/README.md
payneBrandon: It does! It uses a 3rd party SNMP library to do it, and has a few issues. I don't think it's set up to push MAP messages, but that would be a good place for us to put that functionality to keep it centralized.

rsu_automation/src/snmpwalk.py
```diff
@@ -0,0 +1,113 @@
...
+	# no message forwarding configurations, or that some form error occurred in 
+	# reading an RSU's SNMP configuration data. In either scenario, simply returning an 
+	# empty response will suffice for the first implementation.
+	if len(output) < 10:
+		return {}
+
+	snmp_config = {}
+
+	for line in output:
+		prop, raw_value = line.strip().split(' = ')
```
payneBrandon, line 106: We talked about this yesterday,  but wanted to pop it in here as well. Let's get a few comments in this section so we know what is supposed to be happening in here.

payneBrandon (changes requested):
Just the one comment, otherwise looking good!

rsu_automation/src/snmpwalk.py
drewjj: addressed

rsu_automation/README.md
drewjj: it is available somewhere for free, I have no idea where though

rsu_automation/Dockerfile
drewjj: Addressed and tested

rsu_automation/README.md
drewjj: addressed

payneBrandon (approved):
Looks good!

PR #36, opened 2022-02-02
Fixed missing zabbix data bug

mwodahl:
Added environment variables to App.js
Created functions in Map.js to handle errors when RSU data is missing to prevent app from crashing.


cdot-web-app/src/App.js
```diff
@@ -28,9 +30,9 @@ class App extends Component {
 
   async getRsuData() {
     try {
-      let rsuInfoData = await axios.get('https://rsu-manager-apigateway-5xm3e3o5.uc.gateway.dev/rsuinfo');
-      let rsuOnlineData = await axios.get('https://rsu-manager-apigateway-5xm3e3o5.uc.gateway.dev/rsu-online-status');
-      let rsuCountsData = await axios.get('https://rsu-manager-apigateway-5xm3e3o5.uc.gateway.dev/rsucounts');
+      let rsuInfoData = await axios.get(EnvironmentVars.getBaseApiUrl() + '/rsuinfo');
```
payneBrandon, line 33: The "/rsuinfo" portion is already handled if we instead use `EnvironmentVars.rsuInfo`. Same can be applied to below endpoints.

cdot-web-app/src/components/Map.js
```diff
@@ -29,6 +29,23 @@ function Map(props) {
...
+  function isOnline() {
+    try {
+      return props.rsuOnlineStatus[selectedRsu.properties.Ipv4Address].hasOwnProperty("last_online") ? 
+      props.rsuOnlineStatus[selectedRsu.properties.Ipv4Address].last_online + " UTC" : "No Data"
+    } catch (error) {
+      return "No Data"
+    }
+  }
+
+  function getStatus() {
```
payneBrandon, line 41: Rather than relying on try/catch, which is costly when the error is thrown, let's add a check for the property to exist in the dictionary:

```
function getStatus() {
    return (selectedRsu.properties.Ipv4Address in props.rsuOnlineStatus) ?
      props.rsuOnlineStatus[selectedRsu.properties.Ipv4Address].current_status : "Offline";
  }
```

cdot-web-app/src/components/Map.js
```diff
@@ -29,6 +29,23 @@ function Map(props) {
     }
   }, []);
 
+  function isOnline() {
```
payneBrandon, line 32: Same as below, let's re-write this to not rely on a try/catch but instead to check for the property existence before use:

```
function isOnline() {
    return (selectedRsu.properties.Ipv4Address in props.rsuOnlineStatus
      && props.rsuOnlineStatus[selectedRsu.properties.Ipv4Address].hasOwnProperty("last_online")) ?
      props.rsuOnlineStatus[selectedRsu.properties.Ipv4Address].last_online : "No Data";
  }
```

payneBrandon (changes requested):
A couple comments, but looking good!

mwodahl:
Updated App.js to use environment vars. Replaced try/catch blocks to check for existence instead in map.js

payneBrandon (approved):
Looks good to me!

PR #41, opened 2022-03-14
React updates for RSU Manager Authorization

drewjj:
This pull request includes:
- New Google OAuth login support with communication to back-end cloud functions
- New RSU REST API command page to execute commands from
  - Only appears as navigable to after the user has logged in to the RSU Manager
  - Role permissions change how the menu is displayed, only displaying feature the authorized user would be allowed to run. The actual authorization occurs in the back-end still. This is just a UI thing.
- New environment variables for the new endpoints
- Rework of the BSM Count side panel toggle visibility button to an 'X' button that stays in the scope of the side panel's React component


drewjj:
Good to be reviewed now @payneBrandon

cdot-web-app/src/App.js
```diff
@@ -1,60 +1,137 @@
...
+            "key": key,
+            "rsu": key,
+            "road": value.road,
+            "count": value.count
+          };
+        });
+        this.setState({ 'countList': workingList });
+      } catch (exception_var) {
+        console.error(exception_var);
+        this.setState({ 'loading': false });
```
payneBrandon, line 91: Rather than setting the loading false in this catch and below, let's add it to a "finally" block on the outside try/catch layer. That way we are always guaranteed to set loading to false regardless of exceptions.

cdot-web-app/src/App.js
```diff
@@ -1,60 +1,137 @@
...
-          "count": value.count
-        };
-      });
-
-      this.setState({ 'countList': workingList });
+      this.setState({ 'loading': true });
+
+      var authid = ''
+      if (this.state.authLoginData && 'token' in this.state.authLoginData)
+        authid = this.state.authLoginData.token
```
payneBrandon, line 44: Should we add an "else" case here and display an error message to the user that login failed?

payneBrandon: Actually, what's the difference between this state.authLoginData.token and the token we pass into this function? Can we just keep using that token value since we check it for null above?

cdot-web-app/.env.development
```diff
@@ -1,4 +1,5 @@
 BUILD_PATH='./build-dev'
```
payneBrandon, line 1: Do we need to update the `sample.env.local` file with these changes as well?

cdot-web-app/src/App.js
```diff
@@ -1,60 +1,137 @@
...
+    this.state = {
+      selectedRsu: null,
+      rsuData: [],
+      rsuOnlineStatus: {},
+      rsuCounts: {},
+      countList: [],
+      currentSort: "",
+      loading: this.checkLoginData(),
+      expired: false,
+      showSideBar: true,
```
payneBrandon, line 27: This `showSideBar` seems to be unnecessary at this point since the menu handles its own visibility at this point.

payneBrandon (commented):
Just a note, I think we'll want to lock this down just a bit further to prevent access from anything but the login page to all unauthenticated users (see https://stackoverflow.com/questions/47476186/when-user-is-not-logged-in-redirect-to-login-reactjs for some discussion on how to do this with routing). Currently I can manipulate the state variables through the console to gain access to the mapping page. I can't see any data since we include the JWT with all our calls, but I can get further than comfortable with this.

cdot-web-app/src/components/Header.js
```diff
@@ -1,46 +1,105 @@
...
+                  render={renderProps => (
+                    <button onClick={renderProps.onClick} id="googlebtn">Login with Google</button>
+                  )}
+                  clientId={EnvironmentVars.GOOGLE_CLIENT_ID}
+                  onSuccess={this.handleLogin}
+                  onFailure={this.handleFailure}
+                  cookiePolicy={'single_host_origin'}
+                ></GoogleLogin>
+                { this.state.loginFailure && (
+                  <h3 id="loginMessage">User Unauthorized</h3>)}
```
payneBrandon, line 93: Should we add a "contact support" or "request access" link here?

cdot-web-app/src/components/Header.js
```diff
@@ -1,46 +1,105 @@
...
+  handleFailure = (result) => {
+    alert(result);
+  }
 
-const gridStyle = {
-  alignment: "center"
-}
+  handleLogin = async (googleData) => {
+    this.props.setExpired(false);
+    const res = await fetch(EnvironmentVars.googleAuth, {
```
payneBrandon, line 25: There's a couple of cases we may want to explicitly handle here. For instance a 401 unauthorized we could check for and return pertinent info to a user (ask them to request access, or contact support, etc.). Also, if we ran into a 404 or 500 error we may want to fire off an email to support as something has gone wrong.

cdot-web-app/src/components/Menu.js
```diff
@@ -1,34 +1,56 @@
...
-            <div className="header">
-              <div onClick={() => this.props.sortBy('rsu')}>RSU</div>
-              <div onClick={() => this.props.sortBy('road')}>Road</div>
-              <div onClick={() => this.props.sortBy('count')}>Count</div>
-            </div>
-            <div className="body">
-              {rows}
+      <div>
+        { this.state.display ? (
+          <div style={this.menuStyle} id="sideBarBlock" class="visibleProp">
```
payneBrandon, line 34: I think these "class" values need to be changed to "className".

cdot-web-app/src/components/RsuRebootMenu.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+    }
+  }
+
+  render() {
+    return (
+      <div id="snmpdiv">
+        <h2 id="snmpheader">Administrator Reboot</h2>
+
+        <button id="refreshbtn" onClick={this.handleReboot}>Reboot</button>
```
payneBrandon, line 49: For the reboot button, let's add a confirm "are you sure?" type message in there so we don't have anyone fat-finger it and reboot a machine.

cdot-web-app/src/pages/Configure.js
```diff
@@ -0,0 +1,72 @@
...
+
+  handleSnmpMsgFwdRefresh = (msgFwdConfig) => {
+    this.setState({ 'msgFwdConfig': msgFwdConfig });
+  };
+
+  render() {
+    return (
+      <div style={windowStyle}>
+          {
+            this.props.selectedRsu ? (
```
payneBrandon, line 22: I like checking the props and allowing the user to select from the previous page. I think we'll want to allow them to select an RSU from here as well though. 

Maybe we can swap it around so the user opens an RSU menu and clicks on an "edit" link to take them to this page. If the user selects this tab and hasn't clicked an "edit" button we could have a dropdown type-ahead style box to select an RSU here as well. That way they can select it geospatially or by IP to allow maximum flexibility.

cdot-web-app/src/pages/Configure.js
```diff
@@ -0,0 +1,72 @@
...
+                <div>
+                  <SnmpwalkMenu 
+                    msgFwdConfig={this.state.msgFwdConfig}
+                    handleSnmpMsgFwdRefresh={this.handleSnmpMsgFwdRefresh}
+                    isLoginActive={this.props.isLoginActive}
+                    setLoading={this.props.setLoading}
+                    token={this.props.authLoginData.token}
+                    ip={this.props.selectedRsu.properties.Ipv4Address}
+                    road={this.props.selectedRsu.properties.PrimaryRoute}/>
+                  { ['admin', 'operator'].includes(this.props.authLoginData.data.role) && (
```
payneBrandon, line 34: Let's abstract this role check out to a function so we can change these roles easier in the future. We will likely want to make that a check with our service for the roles as well - I can currently manipulate the state to show/hide admin controls by changing my "role" value.

cdot-web-app/src/pages/Map.js
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from "react"
 import ReactMapGL, { Marker, Popup } from "react-map-gl"
-import RsuMarker from './RsuMarker';
+import RsuMarker from '../components/RsuMarker';
```
payneBrandon, line 3: I appreciate the new organization here!

cdot-web-app/src/components/SnmpItem.js
```diff
@@ -0,0 +1,49 @@
+import React, { Component } from 'react';
```
payneBrandon, line 1: Let's rename this file to correspond with the class name, SnmpwalkItem

cdot-web-app/src/components/SnmpwalkMenu.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react';
+import EnvironmentVars from './../EnvironmentVars';
+import SnmpItem from './SnmpItem';
+
+import './css/SnmpwalkMenu.css';
+
+class SnmpwalkMenu extends Component {
+  constructor(props) {
```
payneBrandon, line 8: VSCode tells me we can get rid of this constructor since it isn't doing anything but calling the super.

payneBrandon: Maybe on load we should trigger a refresh if an RSU is selected?

cdot-web-app/src/components/Header.js
drewjj: I never thought this was in scope of the security rework

cdot-web-app/src/components/Header.js
drewjj: This sounds very reasonable. This would be some work, however, are we wanting to make new user stories for this or is this a need for this pr?

cdot-web-app/src/components/Header.js
drewjj: Definitely can add this. We also need the help page implemented at some point.

cdot-web-app/src/components/Menu.js
drewjj: do they? I am applying CSS using these. I am no js/css expert though

cdot-web-app/src/components/SnmpwalkMenu.js
drewjj: > Maybe on load we should trigger a refresh if an RSU is selected?

Not sure what you mean by this

cdot-web-app/src/pages/Configure.js
drewjj: Show not keep role as a state?

cdot-web-app/src/pages/Configure.js
drewjj: I think this makes sense. I was definitely not intending this to remain this way since I also know we want to eventually deploy SNMP configurations to multiple RSUs at once. This menu isn't too allowing of that.

cdot-web-app/src/App.js
drewjj: Addressed

cdot-web-app/src/App.js
```diff
@@ -1,60 +1,137 @@
...
+      authLoginData: this.checkLoginData() ? JSON.parse(localStorage.getItem('authLoginData')) : null,
+    }
+
+    var tid = this.checkLoginData() ? JSON.parse(localStorage.getItem('authLoginData')) : null;
```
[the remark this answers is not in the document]
drewjj: This one is tough, if we do this the state doesn't actually get set in time and we always have a null value.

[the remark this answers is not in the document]
drewjj: eh, nvm. it seems to work now. I thought I tried this in the past and I had an issue with this. It is addressed.

cdot-web-app/src/App.js
drewjj: None. I streamlined this. Addressed.

cdot-web-app/.env.development
drewjj: I missed this file. Addressed.

cdot-web-app/src/App.js
drewjj: Addressed.

cdot-web-app/src/components/RsuRebootMenu.js
drewjj: Looking at adding a pop up for this pull request for this

cdot-web-app/src/components/Menu.js
drewjj: Addressed

cdot-web-app/src/components/SnmpItem.js
drewjj: Addressed

drewjj:
Looks like we have a number of task items here that I will have a hard time fitting into my current sprint so I am looking at making backlog tickets for them. Here are the items:

- Lock down the web application further
  - Separate login menu from the application page by separating the classes
  - Remove the role state so it cannot be altered and request role validation on menu configuration menu load
- Contact support screen
- Help screen
- Better robust handling of different types of login failures 
  - Display more useful error data dependent on error code
  - Send an email to devs when failures occur
- Rework Configuration menu
  - Allow for mass configuration
  - Not require a RSU to be selected but maintain selected RSU if one is selected prior to switching to this menu

Let me know what you think.

cdot-web-app/src/components/SnmpwalkMenu.js
drewjj: > VSCode tells me we can get rid of this constructor since it isn't doing anything but calling the super.

Addressed

cdot-web-app/src/pages/Configure.js
drewjj: Temporarily addressed in comment below

cdot-web-app/src/pages/Configure.js
drewjj: Temporarily addressed in comment below

cdot-web-app/src/components/Header.js
drewjj: Temporarily addressed in comment below

cdot-web-app/src/components/Header.js
drewjj: Temporarily addressed in comment below

cdot-web-app/src/components/RsuRebootMenu.js
drewjj: Addressed. Styling isn't perfectly matching however I think this will all probably change soon so I wasn't too worried about it. Let me know if you want me to do something about that for this pull request.

drewjj:
Pull request is ready to be re-reviewed @payneBrandon

payneBrandon:
> Looks like we have a number of task items here that I will have a hard time fitting into my current sprint so I am looking at making backlog tickets for them. Here are the items:
> 
> * Lock down the web application further
>   
>   * Separate login menu from the application page by separating the classes
>   * Remove the role state so it cannot be altered and request role validation on menu configuration menu load
> * Contact support screen
> * Help screen
> * Better robust handling of different types of login failures
>   
>   * Display more useful error data dependent on error code
>   * Send an email to devs when failures occur
> * Rework Configuration menu
>   
>   * Allow for mass configuration
>   * Not require a RSU to be selected but maintain selected RSU if one is selected prior to switching to this menu
> 
> Let me know what you think.

Those all seem reasonable to add as backlog tasks to me. Let me know if you want to get those added or if you want me to.

cdot-web-app/src/components/SnmpwalkMenu.js
payneBrandon: I was just meaning that when a user accesses the configuration page and this SnmpWalkMenu loads we may want to immediately reach out and pull configuration info if they've selected an RSU so they don't have to click again to get data. 

We can work that into the other RSU selection changes with future backlog items though.

cdot-web-app/src/pages/Configure.js
payneBrandon: We can discuss keeping the role in state or fetching from a service when we write up the task mentioned below. If we can secure the state such that a user cannot manipulate their roles to view additional data, I'm all for keeping them in state.

payneBrandon (approved):
This looks good to me! Just let me know if you are going to write up those backlog tasks or if I need to hit them up. Thanks!

PR #42, opened 2022-03-29
RSU Manager Security Rework

drewjj:
This pull request was going to include all of the changes but since there are already existing pull requests I have the updates to the React app in the already existing React app pull request. I have the updates to the RSU REST Auth cloud function in that pull request.

This pull request includes a rework of the API Gateway into a cloud function. This also reworks the rsu-google-auth cloud function into the gateway cloud function instead of keeping it as a separate cloud function since user auth is a required part of the security rework.

I can upload a diagram for the security workflow if this is at all confusing.


GCP_cloud_functions/rsu-manager/rsu-gateway/main.py
```diff
@@ -0,0 +1,106 @@
...
+  print(f'Received request for {request.path}')
+  # Verify requested resource/endpoint is supported
+  if request.path not in resources:
+    return ('Endpoint unknown', 400, {})
+
+  # If CORS request, handle it according to requested resource/endpoint
+  if request.method == 'OPTIONS':
+    _headers = {
+      'Access-Control-Allow-Origin': '*',
+      'Access-Control-Allow-Methods': resources[request.path][0],
```
payneBrandon, line 37: Clever!

GCP_cloud_functions/rsu-manager/rsu-gateway/main.py
```diff
@@ -0,0 +1,106 @@
...
+    # Parse any HTTP arguments out of URL to pass on
+    args = request.full_path.split('?')[1]
+    if args != '':
+      args = f'?{args}'
+    full_url = f'{endpoint}{args}'
+
+    print(f'Hitting following URL: {full_url}')
+    r = requests.get(full_url, headers=h)
+
+  elif resources[request.path][0] == 'POST':
```
payneBrandon, line 95: We'll want an "else" case in here as well just in case someone tries to call with a "PUT" or "DELETE". Otherwise this will throw an exception when we try to return the "r.text" below.

payneBrandon (changes requested):
Looking pretty solid here! We'll also want to update the README to reflect the change away from API Gateway in favor of this new solution.

GCP_cloud_functions/rsu-manager/rsu-gateway/main.py
drewjj: Actually that should be caught on line 49 and anything not supported will be returned as an unsupported method type. (CORS will also catch this if this is attempted on a browser)

I can still add an else I suppose but it would actually never get run unless someone added a new endpoint for a method type that wasn't defined here at some point down the road.

drewjj: Also, the reason why I didn't just make it as an else to return an unsupported method is because I wanted to have control over methods that may allow GETs, but not POSTs.

drewjj: As I think about it more, you could technically have two conditions that would combine the two - "if supported AND is a GET". But there's a lot of ways to refactor the same logic. If the way I implemented it is confusing I can do it this way too. I just wanted to handle the authorization before this step which I only wanted to handle in the scenario the request was supported first.

payneBrandon: This makes sense, I just missed the check for supported verbs above.

payneBrandon:
> > Looking pretty solid here! We'll also want to update the README to reflect the change away from API Gateway in favor of this new solution.
> 
> Will do! Will add that with the diagram.

Will this be a part of this PR or no?

drewjj:
> > > Looking pretty solid here! We'll also want to update the README to reflect the change away from API Gateway in favor of this new solution.
> > 
> > 
> > Will do! Will add that with the diagram.
> 
> Will this be a part of this PR or no?

Yes! I just added it to the pull request.

payneBrandon (approved):
Looks good to me!

PR #43, opened 2022-04-04
Update RSU-Manager to show dates in MST

mwodahl:
Added date formatting for rsu-online-status to show last online time in MST, and updated rsu-count-emailer to only show date and not time.


GCP_cloud_functions/rsu-manager/rsu-count-emailer/main.py
```diff
@@ -80,9 +80,9 @@ def create_count_table(start, end):
   return html
 
 def create_email_content():
-  start = datetime.strftime(datetime.now() - timedelta(1), '%Y-%m-%d 00:00:00')
-  end = datetime.strftime(datetime.now(), '%Y-%m-%d 00:00:00')
-  html = f'<h2>{str(os.environ["PROJ_NAME"]).upper()} GCP Count Report {start} UTC - {end} UTC</h2>' \
+  start = datetime.strftime(datetime.now() - timedelta(1), '%Y-%m-%d')
```
payneBrandon, line 83: This is actually going to be slightly more complicated than we're making it out here. Data in our database is stored in UTC time, but we want it displayed in current mountain time (MST or MDT depending on if we're in daylight savings or not). This means that we need to use the start and end times here in mountain time in the html, which we've already done. The trick then is to translate those mountain time values into UTC for the actual query. Let me know when you get in if you want to discuss.

GCP_cloud_functions/rsu-manager/rsu-online-status/main.py
```diff
@@ -116,6 +117,12 @@ def get_last_online_data():
   
   return result
 
+def convertDate(d):
```
payneBrandon, line 120: Similar to above, this convertDate function we will want to take the database stored UTC time and return mountain time for display

payneBrandon (changes requested):
A couple issues as commented. Let me know when you get in if you want to discuss.

GCP_cloud_functions/rsu-manager/rsu-count-emailer/main.py
mwodahl: Added functions to format the start/end dates in mountain time for the html and utc time for the query.

GCP_cloud_functions/rsu-manager/rsu-online-status/main.py
mwodahl: Changed convertDate function to return mountain time for display

mwodahl:
Resolved merge conflict with main branch and updated functions to return correct date time zone information

PR #45, opened 2022-04-07
Rsu-manager count update

mwodahl:
Allows user to toggle between different message types + updates table and map accordingly


payneBrandon (commented):
node_modules shouldn't be check into source control, let's make sure there's a .gitignore in there to prevent that and remove.

payneBrandon:
Looks like we also have some merge conflicts now that the other branch has been merged in.

payneBrandon (changes requested):
A few changes as mentioned, tabling this until we get that function running properly

mwodahl:
removed node_modules, fixed merge conflicts with main branch

PR #46, opened 2022-04-08
DevOps pipeline update for RSU Manager

drewjj:
Includes pipeline support for:
- New cloud functions
- RSU Manager new environment variables


cdot-web-app/src/EnvironmentVars.js
```diff
@@ -9,14 +9,24 @@ class EnvironmentVars {
...
+
     static rsuInfo = `${this.getBaseApiUrl()}/rsuinfo`;
     static rsuOnline = `${this.getBaseApiUrl()}/rsu-online-status`;
     static rsuCounts = `${this.getBaseApiUrl()}/rsucounts`;
     static googleAuth = `${this.getBaseApiUrl()}/rsu-google-auth`;
     static rsuRestAuth = `${this.getBaseApiUrl()}/rsu-rest-auth`;
 
     static MAPBOX_TOKEN = process.env.REACT_APP_MAPBOX_TOKEN;
-    static GOOGLE_CLIENT_ID = process.env.REACT_APP_GOOGLE_CLIENT_ID;
+    static GOOGLE_CLIENT_ID = this.getClientId();
```
payneBrandon, line 29: Was this not switching based on environment already?

payneBrandon (approved):
Looks good to me, just the one comment!

cdot-web-app/src/EnvironmentVars.js
drewjj: No, not really. The idea was that each environment would use the same environment variable name but with a different value. What I am noticing in the RSU Manager pipeline is that there pretty much needs to be separate variables for each environment similar to the RSU Gateway ().

payneBrandon: I wonder if we're just not properly building for each environment then. I wouldn't think we'd have to have separate variable names for each environment like that. We'll want to check our ci/cd pipelines and make sure we're specifying the environment when building each.

payneBrandon: Are you using the build and release pipelines to deploy or doing this manually? The build pipeline looks like it should be creating appropriate environment-specific builds running the `env-cmd -f <filename>...` syntax - if that's not working we definitely need to dig into why and how to get those builds working properly.

payneBrandon (approved):
Looks good to me!

PR #48, opened 2022-04-14
Rsu manager querycounts update

mwodahl:
Update main.py to allow for date/time testing


GCP_cloud_functions/rsu-manager/rsu-querycounts/main.py
```diff
@@ -1,34 +1,40 @@
 from google.cloud import bigquery
 from datetime import datetime, timedelta
+from dateutil.parser import parse
+import pytz
 import os
 
-def query_rsu_counts(message_type):
+def query_rsu_counts(message_type, userDates):
```
payneBrandon, line 7: Rather than passing in "userDates" as an array, let's just pass in "startDate" and "endDate". That way we can easily know which date represents what, and we can apply validation easily for required fields.

payneBrandon (changes requested):
Just the one comment, looking good!

GCP_cloud_functions/rsu-manager/rsu-querycounts/main.py
mwodahl: Replaced userDates with start and end date args

payneBrandon (approved):
Looks good to me!

GCP_cloud_functions/rsu-manager/rsu-querycounts/main.py
```diff
@@ -40,9 +46,28 @@ def entry(request):
...
+    startDate = request.args.get('start', default=((datetime.now() - timedelta(1)).strftime("%Y-%m-%dT%H:%M:%S")))
+    endDate = request.args.get('end', default=((datetime.now()).strftime("%Y-%m-%dT%H:%M:%S")))
+    msgList = ['SSM', 'BSM', 'SPAT', 'SRM', 'MAP']
+    if msgType.upper() not in msgList:
+      return ("Invalid Message Type.\nValid message types: SSM, BSM, SPAT, SRM, MAP", 200, headers)
+    try:
+      parse("\'" + startDate + "\'")
+      parse("\'" + endDate + "\'")
+    except:
+      return ("Date format: YYYY-MM-DDTHH:MM:SS", 200, headers)
```
drewjj, line 65: When we have an exception like this, we should return a response code that is not 200 since this represents the function ran right. A code like 412 might work better since this is checking the condition/content of the headers before running the function.

GCP_cloud_functions/rsu-manager/rsu-querycounts/main.py
```diff
@@ -1,34 +1,40 @@
...
+      item = {
+        "road": row["Road"],
+        "date": datetime.strftime(row["Date"], '%Y-%m-%dT00:00:00'),
+        "count": row["Count"]
+      }
       result[row["RSU"]] = item
 
     return result
 
+def formatDate(d):
```
drewjj, line 34: Pythonic nitpick, this function should go above the query_rsu_counts function since it is called by it. Python expects all code that is called to be organized this way and thereby it is easier to read for Python scripts. If the main function wasn't properly at the bottom of the script this might actually fail to run otherwise!

That's at least how Python worked the last time I checked.

drewjj (changes requested):
Just a couple things. Functionality of the script looks great.

PR #49, opened 2022-04-18
Hide help button for deployment

drewjj:
This pull request just comments out the help button from the header so it will not display since it is not functioning properly when deployed.


payneBrandon (approved):
Looks good to me

PR #53, opened 2022-04-19
Bug/extra-data

payneBrandon:
Removing extra data from call to get_ping_data


drewjj (approved):
Very cool. That debugging stuff is super neat.

